PR #8, opened 2018-01-28
Improved graphql factory

partyka95:
Hey. When the "resolvers" property was passed to the "createSchema" function, it was overwritten and I could not add other types (for example, the scalar types for dates - https://www.apollographql.com/docs/graphql-tools/scalars.html).

### DIFF
--- a/lib/graphql.factory.ts
+++ b/lib/graphql.factory.ts
@@ -18,11 +18,14 @@ export class GraphQLFactory {
   ) {}
 
   createSchema(
-    schemaDefintion: IExecutableSchemaDefinition = { typeDefs: {} },
+    schemaDefintion: IExecutableSchemaDefinition = { typeDefs: [] },
   ) {
     return makeExecutableSchema({
       ...schemaDefintion,
-      resolvers: this.resolversExplorerService.explore(),
+      resolvers: {
+        ...this.resolversExplorerService.explore(),
+        ...(schemaDefintion.resolvers || {}),
+      },
     });
   }
 


4F2E4A2E:
ping

rgolea:
Anything on this PR? I would like to create my own scalar types and I would like to use this as soon as possible.

kamilmysliwiec:
Thanks! Merged 🙂

rgolea:
@kamilmysliwiec could we add decorators in order to use this?
I'm thinking that something like:
```
scalar Date
```

```
import { ScalarResolver, Scalar } from '@nestjs/graphql';

@ScalarResolver()
export class DateType implements Scalar{
   public parseValue(value) {
      return new Date(value); // value from the client
   }
   public serialize(value) {
      return value.getTime(); // value sent to the client
   }
   parseLiteral(ast) {
      if (ast.kind === Kind.INT) {
        return parseInt(ast.value, 10); // ast value is always in string format
      }
      return null;
   }
}
```
Example from the [apollo documentation](https://www.apollographql.com/docs/graphql-tools/scalars.html#Date-as-a-scalar)

kamilmysliwiec:
Looks pretty amazing. Do you want to try to implement this? You could follow currently available decorators 🔥

rgolea:
@kamilmysliwiec I will try to... Could you not publish v3 until I do? I would need some help with it...

rgolea:
Just reviewed the code. With `@Resolver()` you can do basically the same. So...
```
scalar Date
```
```
import { Resolver } from '@nestjs/graphql';

@Resolver('Date')
export class DateResolver {
   public parseValue(value) {
      return new Date(value); // value from the client
   }
   public serialize(value) {
      return value.getTime(); // value sent to the client
   }
   parseLiteral(ast) {
      if (ast.kind === Kind.INT) {
        return parseInt(ast.value, 10); // ast value is always in string format
      }
      return null;
   }
}
```

kamilmysliwiec:
I didn't even realize that 😅 @rgolea, do you want to create a Pull Request to the docs? 🎉

triedal:
@rgolea when I use this method my time parameter is a pending promise. Any idea what I could be doing wrong?

rgolea:
do `await input.time`? for some reason I couldn't figure out, it just... gives a promise from the scalar...

kamilmysliwiec:
yeah, that's true. Nest creates an evaluation context for each resolver/mutation/query which is asynchronous by default. I'll provide a dedicated solution proposed by @rgolea, `@ScalarResolver()` or just `@Scalar()` soon

rgolea:
@kamilmysliwiec have you managed to do a video tutorial on the structure of the project so we can help more on the project?

triedal:
@kamilmysliwiec awesome, looking forward to it! @rgolea I was able to accomplish what I'm looking for by doing `input.time = await input.time;`. This seemed a bit hacky so I wasn't sure if there was a better solution. This will work in the interim though.

triedal:
btw, slightly off-topic but @kamilmysliwiec this framework is outstanding. It's everything I've been looking for in a node framework. Thank you so much for taking the time to put this together.

PR #22, opened 2018-05-19
Fixed undefined instance error when combined with Angular Universal

bojidaryovchev:
I integrated grapql in my project (https://github.com/bojidaryovchev/nest-angular/tree/master/src/server/modules) following this example https://github.com/nestjs/nest/tree/master/sample/12-graphql-apollo.. But I also have Angular Universal from the angular universal example.. And I found out that when we have both our GraphqlModule and AngularUniversalModule, an undefined instance appears in the modules instances in the explore and flapMap functions in resolvers-explorer.service.. Follow the images for further information:

https://ibb.co/hhX5RT
https://ibb.co/fujKmT
https://ibb.co/cafs6T

I found this fix to make things work however I am not completely sure that it is sufficient.. It seems to be, but I cannot determine.. Anyway, I shall make a pull request

### DIFF
--- a/lib/resolvers-explorer.service.ts
+++ b/lib/resolvers-explorer.service.ts
@@ -32,7 +32,7 @@ export class ResolversExplorerService {
     );
     const resolvers = this.flatMap(modules, instance =>
       this.filterResolvers(instance),
-    );
+    ).filter(r => r);
     return this.groupMetadata(resolvers);
   }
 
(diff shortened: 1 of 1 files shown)


matcic:
@kamilmysliwiec something similar happened to me as well while trying to apply some permission policies to my Graphql resolvers based on custom decorators. The thing is that I believe [here](https://github.com/nestjs/graphql/blob/master/lib/services/base-explorer.service.ts#L24) the instance property is undefined for any injected instance that is defined as a provider belonging to the current module (which has not been instantiated yet). A simple `filter(value => value.instance)` (similar to what @bojidaryovchev suggested) fixed for me. Please let us know if this looks like a reasonable fix 😄

marcus-sa:
You should separate your SSR from your API

PR #81, opened 2018-10-02
fix(GraphQLFactory): avoid unnecessary overwrites

soroushj:
Do not overwrite the TS file if it already exists with the same text (source). This prevents restart loops with tools like `nodemon`.

## PR Checklist

## PR Type
A minor change in behavior

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe: A minor change in behavior
```

The output TS file is always overwritten.

Issue Number: N/A

## What is the new behavior?
The output TS file is overwritten only if the text (source) has changed.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/lib/graphql.factory.ts
+++ b/lib/graphql.factory.ts
@@ -1,3 +1,4 @@
+import { existsSync, lstatSync, readFileSync } from 'fs';
 import { Injectable } from '@nestjs/common';
 import { gql, makeExecutableSchema, mergeSchemas } from 'apollo-server-express';
 import { MergeInfo } from 'graphql-tools/dist/Interfaces';
@@ -75,6 +76,12 @@ export class GraphQLFactory {
       options.definitions.path,
       options.definitions.outputAs,
     );
-    await tsFile.save();
+    if (
+      !existsSync(options.definitions.path) ||
+      !lstatSync(options.definitions.path).isFile() ||
+      readFileSync(options.definitions.path, 'utf8') !== tsFile.getText()
+    ) {
+      await tsFile.save();
+    }
   }
 }


kamilmysliwiec:
LGTM, thanks @soroushj

val-samonte:
@soroushj @kamilmysliwiec 
Sorry i don't know where to put my comment, but i found out that comparing file changes is not accurate. The comments were present on readFileSync and were absent in tsFile.getText()

soroushj:
@val-samonte could you please include the steps required to reproduce the bug?

val-samonte:
I am not really sure maybe because i am using angular nx + nestjs has something to do with this, but following the example source (https://github.com/nestjs/nest/tree/master/sample/12-graphql-apollo/src) adding a comment inside the graphql.schema.ts (eg. /* tslint: disabled */) causes a loop when i do "ng serve nestjs-app".

Sorry if I can't give a proper steps, but I did debug the graphql.resolve file and found out that the readFileSync and tsFile.getText() return different outputs (one has included the comments, one hasn't).

dantman:
I'm having the same issue. Just started setting up GraphQL in my project and getting an infinite loop because of a `/* tslint:disable */` comment when I don't even use tslint (I use eslint instead).

`tsFile.save()` saves text with the tslint comment. `tsFile.getText()` returns text without the comment.

1. You start with a blank schema file and the app starts up
2. It generates a file with a tslint comment
3. App restarts due to file creation (good)
4. `tsFile.getText()` text without tslint comment is compared to file content with tslint comment and fails
5. `tsFile.save()` saves the same file with tslint comment
6. App restarts due to file update (bad)
7. Steps 4-6 continually loop over and over because getText always returns without the comment and save always saves with the comment
8. You manually delete the tslint comment from the file
9. `tsFile.getText()` text without tslint comment equals your modified file and save is not called
10. You edit one of your .graphql schema files and do something that triggers a restart (good)
11. `tsFile.getText()` text without tslint comment for new schema is compared to your file content for old schema and does not equal (good)
12. `tsFile.save()` saves the same file with tslint comment again
13. The 7. infinite loop returns

soroushj:
@dantman, thank you for describing the issue in detail. I'll take a look into it.

soroushj:
@dantman, I opened an issue (#136) and submitted a pull request (#137) that fixes the problem. Thank you for reporting the bug.

PR #114, opened 2019-01-21
bugfix() Sort lookup parameters

simhnna:
Required lookup parameters were not generated at the beginning of the
methods. This caused the generated files to be invalid.
This change makes sure that all required fields are at the beginning of
the parameter list followed by all optional parameters.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When generating schema types, the required function parameters were not in the beginning.

## What is the new behavior?
Generated function parameters are now sorted based on them being required and then based on the name.

```
[x] Yes
[ ] No
```

## Other information

### DIFF
--- a/lib/graphql-ast.explorer.ts
+++ b/lib/graphql-ast.explorer.ts
@@ -154,6 +154,13 @@ export class GraphQLAstExplorer {
         return this.lookupField(item, parentRef, mode);
     }
   }
+  
+  private sortGraphqlArguments(first: any, second:any): number {
+    if (first.hasQuestionToken && second.hasQuestionToken || !(first.hasQuestionToken || second.hasQuestionToken)) {
+      return first.name === second.name : 0 ? first.name > second.name ? 1 : -1;
+    }
+    return first.hasQuestionToken ? 1 : -1;
+  }
 
   lookupField(
     item: FieldDefinitionNode | InputValueDefinitionNode,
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Sounds interesting. Thanks for the contribution.

lib/graphql-ast.explorer.ts
```diff
@@ -184,7 +184,21 @@ export class GraphQLAstExplorer {
       returnType: `${type} | Promise<${type}>`,
       parameters: this.getFunctionParameters(
         (item as FieldDefinitionNode).arguments,
-      ),
+      ).sort((a, b) => {
```
kamilmysliwiec: perhaps we should extract this comparator to a separate function + simplify it a little bit, for example this
```typescript
else if (a.hasQuestionToken) {
    return 1;
} else {
    return -1;
}
```
to:
```typescript
return a.hasQuestionToken ? 1 : -1;
```

simhnna: Done. I don't have a development environment around now, so I did the changes in github...

lib/graphql-ast.explorer.ts
```diff
@@ -154,6 +154,13 @@ export class GraphQLAstExplorer {
         return this.lookupField(item, parentRef, mode);
     }
   }
+  
+  private sortGraphqlArguments(first: any, second:any): number {
+    if (first.hasQuestionToken && second.hasQuestionToken || !(first.hasQuestionToken || second.hasQuestionToken)) {
+      return first.name === second.name : 0 ? first.name > second.name ? 1 : -1;
```
Mnwa: I think it's not a good practices two conditional operators in one row

PR #260, opened 2019-05-21
Removing  `optional` dependency

aequasi:
## PR Checklist

I'm assuming these are all fine. I just removed a dependency, but it shouldnt require any changes in tests/docs

## PR Type

```
[x] Bugfix
[ ] Feature
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Removed the `optional` package dependency. Was causing issues with ncc/webpack. 
https://github.com/nestjs/nest/issues/2213

Issue Number: 2213

Projects running on Zeit Now will not be able to use graphql + type-graphql, as the require calls don't function properly when transpiled

## What is the new behavior?

Transpiler is now able to statically analyze the require calls to type-graphql

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/lib/decorators/args.decorator.ts
+++ b/lib/decorators/args.decorator.ts
@@ -1,14 +1,17 @@
 import { PipeTransform, Type } from '@nestjs/common';
 import { isObject, isString } from '@nestjs/common/utils/shared.utils';
-import * as optional from 'optional';
 import 'reflect-metadata';
 import { GqlParamtype } from '../enums/gql-paramtype.enum';
 import { BasicOptions } from '../external/type-graphql.types';
 import { lazyMetadataStorage } from '../storages/lazy-metadata.storage';
 import { addPipesMetadata } from './param.utils';
 
-const { Arg: TypeGqlArg, Args: TypeGqlArgs } =
-  optional('type-graphql') || ({} as any);
+let TypeGqlArg, TypeGqlArgs;
+try {
+  const TypeGql = require('type-graphql');
+  TypeGqlArg = TypeGql.Arg;
+  TypeGqlArgs = TypeGql.Args;
+} catch (e) {}
 
 export interface ArgsOptions extends BasicOptions {
   name?: string;
--- a/lib/decorators/mutation.decorator.ts
+++ b/lib/decorators/mutation.decorator.ts
@@ -1,5 +1,4 @@
 import { isString } from '@nestjs/common/utils/shared.utils';
-import * as optional from 'optional';
 import { Resolvers } from '../enums/resolvers.enum';
 import { lazyMetadataStorage } from '../storages/lazy-metadata.storage';
 import {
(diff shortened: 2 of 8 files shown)


gimyboya:
@kamilmysliwiec can you please take a look at this PR [Zeit](http://now.sh/) is a very handy deployment platform.

marcus-sa:
Sorry, but you can't just remove a dependency because your environment doesn't support it.
There's several ways for you to be able to utilize `optional` without actually removing it.
E.g modifying your webpack config would be one way, or you could use `loadPackage` which is exported from `@nestjs/common`.
I can approve that it works, since I'm currently using it in a `@nrwl/workspace` mono repository.

aequasi:
You definitely can, especially if it's an entirely unecessary dependency.

gimyboya:
@marcus-sa I don't see where is the problem in removing the `optional` dependency in this context. when we know it's not going to break things and on top of that, it's actually fixing a bug that is affecting many users preventing them from deploying their Nestjs apps.

marcus-sa:
@gimyboya you do realize `type-graphql` is an optional dependency, right?

aequasi:
And? Please stop being so hostile @marcus-sa . It adds nothing to the conversation here.

gimyboya:
@marcus-sa Sorry, I might not be a seasoned developer as you. but i still want to see this PR go through. 
And as someone who is grateful to this framework and grateful to OSS in general, I only want the best for the community.

So let me try to clarify what i understand so far.
Although `type-graphql` is optional, there are some [edge cases where it's needed](https://github.com/zeit/ncc/issues/377#issuecomment-495563034) and because we are using `optional` in this case, static analysis is failing as you can see in this [issue](https://github.com/zeit/ncc/issues/377) (which is affecting many users including myself).

IMHO i believe that as developers we should strive and cover all edge cases to provide the best development experience. in other words if including `type-graphql` as an optional dependency affects some users (prevents them from deploying to a widely used platform like [now.sh](https://now.sh)) we can always include it to help them OR come up with an elegant solution that if you do have please suggest and provide.

gimyboya:
@kamilmysliwiec a gentle reminder to look at this PR.

kamilmysliwiec:
We can't just remove `optional` dependency because `type-graphql` is actually an optional library not required for at least 50% of `@nestjs/graphql` users. 

In order to support `ncc` and other bundlers, we can use `loadPackage` utility function instead of `optional`. `loadPackage` is using native `require()` underneath, but wraps this call with `try..catch` block so even if someone doesn't have the library installed, it won't break the entire app :) 

Would you like to do it? You can see an example of how to use `loadPackage` in the nestjs/nest repo.

aequasi:
I am not making `type-graphql` non-optional with this PR. This PR just removed the 4-line dependency `optional`. It not a breaking change.

The problem is static analysis. Its far easier for compilers to do static analysis if you are directly calling require, without dynamic arguments.

kamilmysliwiec:
Oh, I can see it now. Could you please resolve conflicts?

kamilmysliwiec:
Merged :) Thanks

PR #277, opened 2019-06-01
fix(@nestjs/graphql): fix array type output of generated properties

dantman:
## PR Checklist

- [n/a] Docs have been added / updated (for bug fixes / features)

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

`prop: [String!]` currently outputs `prop: string[]` because of the `String!`, which is incorrect because whether `prop` is required is based on whether there is a `!` outside the brackets, not inside.

## What is the new behavior?

`prop: [String!]` now outputs `prop?: string[]` instead.

```
[x] Yes
[ ] No
```

Technically this is a breaking change, if there is any `propName: [Foo!]` within a project's graphql schema the outputted definitions will change from `propName: Foo[]` to `propName?: Foo[]`.

"Technically" this update could break something in an edge case were someone has written a bad schema based on this behaviour or . But these edge cases should be simple one-character or arg reordering fixes.

For most people they should just update, let the schema auto-update any `?`s (if there are even any that are affected by this change), and just run their app to make sure it runs. Anything requiring fixes would show up as immediately obvious syntax errors.

It's very unlikely that any methods actually rely on this because the interface method types are incompatible with resolver functions.

## Other information

In query/mutation/subscriber definitions this bug can actually have the result of generating code with an error about required args not being allowed after optional args, even when all your args are optional but an array arg has been incorrectly declared as required.

e.g. `createMessage(authorId: Int!, body: String, attachments: [Attachment!]): Message` will incorrectly result in `createMessage(authorId: number, body?: string, attachments: AttachmentInput[]): Message | Promise<Message>;` instead of `createMessage(authorId: number, body?: string, attachments?: AttachmentInput[]): Message | Promise<Message>;` resulting in a syntax error because `attachments` is declared as required even though in the GQL schema is was declared as optional and in the argument order it's after another optional arg.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ npm-debug.log
 /coverage
 /.nyc_output
 test-schema.graphql
+*.test-definitions.ts
 
 # dist
 /dist
--- a/lib/graphql-ast.explorer.ts
+++ b/lib/graphql-ast.explorer.ts
@@ -200,9 +200,7 @@ export class GraphQLAstExplorer {
 
     const isArray = type.kind === 'ListType';
     if (isArray) {
-      const { required, type: nestedType } = this.getNestedType(
-        get(type, 'type'),
-      );
+      const { type: nestedType } = this.getNestedType(get(type, 'type'));
       type = nestedType;
 
       const typeName = get(type, 'name.value');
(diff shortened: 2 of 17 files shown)


kamilmysliwiec:
Amazing contribution! Thank you for this PR + leaving lots of useful information. I'll publish this as soon as possible since the current behavior is incorrect.

Issue #537, opened 2019-12-28
Bug: iterall not in package.json


andreialecu:
https://github.com/nestjs/graphql/blob/487379b8c0524fe3cdbeb085038d804ba6fff0fd/lib/utils/async-iterator.util.ts#L1

This library imports `iterall` but it is not defined in package.json.

This is a problem with yarn v2, the upcoming version of yarn, resulting in errors such as:

```
Cannot find module 'iterall' from 'async-iterator.util.js'

      at Resolver.resolveModule (../.yarn/cache/jest-resolve-npm-24.9.0-56dbaf2e7d-1.zip/node_modules/jest-resolve/build/index.js:259:17)
      at Object.<anonymous> (../.yarn/virtual/@nestjs-graphql-virtual-9a8fdb1b99/0/cache/@nestjs-graphql-npm-6.5.3-98b2ad3bf3-1.zip/node_modules/@nestjs/graph
ql/dist/utils/async-iterator.util.js:12:19)
```

The library should be added to dependencies.

andreialecu:
This probably worked before because of the behavior described in this comment:

https://github.com/yarnpkg/berry/issues/271#issuecomment-507432524

kamilmysliwiec:
This issue is not related to NestJS

andreialecu:
???

This package is using a library not defined in its package.json. How is this not related?

PR #538, opened 2019-12-28
fix(deps): add iterall dependency

andreialecu:
iterall was imported without being in package.json

This works in some environments because the package steals iterall from its sub dependencies, but it is not a good practice and not compatible with https://yarnpkg.com/lang/en/docs/pnp/ and with upcoming similar features from `npm`

This is a bad practice in general: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md

Fixes https://github.com/nestjs/graphql/issues/537

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "chokidar": "3.3.1",
     "fast-glob": "3.1.1",
     "graphql-tools": "4.0.6",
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Thank you @andreialecu!

Issue #222, opened 2019-04-23
CustomScalar parseLiteral doesn't allow null return type


PachowStudios:
## I'm submitting a...

<pre><code>
[ ] Regression 
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

The samples show that you should return null from parseLiteral if the AST kind doesn't match the expected kind.

```ts
// From https://github.com/nestjs/nest/blob/master/sample/23-type-graphql/src/common/scalars/date.scalar.ts
parseLiteral(ast: any): Date {
  if (ast.kind === Kind.INT) {
    return new Date(ast.value);
  }
  return null;
}
```

Doing this with TS strictNullChecks enabled, results in an error.

## Expected behavior

CustomScalar should allow parseLiteral to return null or undefined as demonstrated by the samples.

## What is the motivation / use case for changing the behavior?

The pattern recommended by the samples should work under TS strict mode.

## Environment
<pre><code>
@nestjs/core: 6.1.1
@nestjs/graphql: 6.0.5
typescript: 3.4.4

SugarDarius:
Hi ! 

I had the same issue with my **tsconfig.json** file and the property _strictNullChecks_ set to **true**.
You can simply do this to resolve your issue : 

```typescript
parseLiteral(ast: any): Date | null {
  return ast.kind === Kind.INT ? new Date(ast.value) : null;
}
```
Because if you check this file : _[https://github.com/nestjs/graphql/blob/master/lib/interfaces/custom-scalar.interface.ts](https://github.com/nestjs/graphql/blob/master/lib/interfaces/custom-scalar.interface.ts)_ you can see that the returned types of the methods of the interfaces are generics : 

```typescript
export interface CustomScalar<T, K> {
  description: string;
  parseValue(value: T): K;
  serialize(value: K): T;
  parseLiteral(ast): K;
}
```

I hope this helps you :)

mateusppereira:
Somehow related do this, eslint rule `@typescript-eslint/no-explicit-any` forces me to specify a type for ast, what should it be?

PR #749, opened 2020-04-05
Fix CustomScalar interface for strictNullChecks

kaorun343:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #222 

In strict mode of TypeScript, the current definition of `CustomScalar` doesn't allow us to return `null` or `undefined`.

## What is the new behavior?

This code will not throw any errors in strict mode.

```typescript
parseLiteral(ast: any): Date {
  if (ast.kind === Kind.INT) {
    return new Date(ast.value);
  }
  return null;
}
```

```
[ ] Yes
[x] No
```

## Other information

References:

- https://github.com/graphql/graphql-js/blob/278bde0a5cd71008452b555065f19dcd1160270a/src/type/definition.d.ts#L293-L296
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html

### DIFF
--- a/lib/interfaces/custom-scalar.interface.ts
+++ b/lib/interfaces/custom-scalar.interface.ts
@@ -1,6 +1,6 @@
 export interface CustomScalar<T, K> {
-  description: string;
-  parseValue(value: T): K;
-  serialize(value: K): T;
-  parseLiteral(ast): K;
+  description: string | null | undefined;
+  parseValue(value: T): K | null | undefined;
+  serialize(value: K): T | null | undefined;
+  parseLiteral(ast): K | null | undefined;
 }
--- a/lib/interfaces/custom-scalar.interface.ts
+++ b/lib/interfaces/custom-scalar.interface.ts
@@ -1,8 +1,12 @@
-import { ValueNode } from 'graphql';
+import {
+  GraphQLScalarLiteralParser,
+  GraphQLScalarSerializer,
+  GraphQLScalarValueParser,
+} from 'graphql';
 
 export interface CustomScalar<T, K> {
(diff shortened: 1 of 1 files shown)


lib/interfaces/custom-scalar.interface.ts
```diff
@@ -1,6 +1,8 @@
+import { ValueNode } from 'graphql';
+
 export interface CustomScalar<T, K> {
-  description: string;
-  parseValue(value: T): K;
-  serialize(value: K): T;
-  parseLiteral(ast): K;
+  description: string | null | undefined;
+  parseValue(value: T): K | null | undefined;
+  serialize(value: K): T | null | undefined;
```
kamilmysliwiec, line 6: Perhaps we should define `Maybe` type instead? Following this implementation https://github.com/graphql/graphql-js/blob/278bde0a5cd71008452b555065f19dcd1160270a/src/type/definition.d.ts#L293-L296

kaorun343: Or how about using these type definitions directly?

https://github.com/graphql/graphql-js/blob/278bde0a5cd71008452b555065f19dcd1160270a/src/type/definition.d.ts#L316-L325

kamilmysliwiec: Good point! I like it

kaorun343:
Could you review this?

kamilmysliwiec:
Thanks!

Issue #775, opened 2020-04-14
mixin type helper


Diluka:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## What is the motivation / use case for changing the behavior?

to create class C with field c and field a in class A and field b in class B
```ts
@ObjectType()
class A {
  @Field()
  a:string;

  @Field()
  a1:string;
}

@InputType()
class B {
  @Field()
  b:string;

  @Field()
  b1:string;
}

@InputType()
class C extends MixinType(PickType(A,['a']),PickType(B,['b']), InputType){
  @Field()
  c:string;
}
```

## Environment

<pre><code>
Nest version: X.Y.Z

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

kamilmysliwiec:
Great idea! Let's track this here https://github.com/nestjs/graphql/pull/776

PR #776, opened 2020-04-14
feat(@nestjs/graphql): add mixin type helper

Diluka:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #775 

## What is the new behavior?
mixin two types by default using the first one's type decorator

```
[ ] Yes
[x] No
```

## Other information


lib/type-helpers/mixin-type.helper.ts
```diff
@@ -0,0 +1,44 @@
+import { Type } from '@nestjs/common';
+import { Field } from '../decorators';
+import { getFieldsAndDecoratorForType } from '../schema-builder/utils/get-fields-and-decorator.util';
+import {
+  ClassDecoratorFactory,
+  inheritValidationMetadata,
+  inheritTransformationMetadata,
+} from './type-helpers.utils';
+
+export function MixinType<A, B>(
```
kamilmysliwiec, line 10: What about using `IntersectionType` instead? We use `mixin` keyword for a different functionality already

lib/type-helpers/mixin-type.helper.ts
```diff
@@ -0,0 +1,44 @@
...
+  classARef: Type<A>,
+  classBRef: Type<B>,
+  decorator?: ClassDecoratorFactory,
+): Type<A> & Type<B> {
```
kamilmysliwiec, line 14: `Type<A & B>`?

lib/type-helpers/mixin-type.helper.ts
Diluka: It is more like `Union` than `Intersection`.

kamilmysliwiec: By the definition of "intersection" in general, you're right. However, in TypeScript, Intersection types are defined as follows:

> An intersection type combines multiple types into one. This allows you to add together existing types to get a single type that has all the features you need. For example, Person & Serializable & Loggable is a Person and Serializable and Loggable. That means an object of this type will have all members of all three types. You will mostly see intersection types used for mixins and other concepts that don’t fit in the classic object-oriented mold. (There are a lot of these in JavaScript!) Here’s a simple example that shows how to create a mixin:

Ref https://www.typescriptlang.org/docs/handbook/advanced-types.html#intersection-types

Diluka: OK. Since TypeScript says `A & B` is intersection type then we use `IntersectionType` as it's name

kamilmysliwiec:
Would you like to create a PR to the docs with this change? Also, we have the same type helpers in the `@nestjs/swagger` package. Would you be interested in creating a PR in this repo as well?

lib/type-helpers/intersection-type.helper.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  fields.forEach((item) => {
+    Field(item.typeFn, { ...item.options })(
+      IntersectionObjectType.prototype,
+      item.name,
+    );
+  });
+
+  Object.defineProperty(IntersectionObjectType, 'name', {
+    value: `Mixin${classARef.name}${classBRef.name}`,
```
Diluka, line 41: Oh, I forgot to rename this one. Is that OK? 

And there is a question about the length of `name`. If someone nested using the helpers, it could be very long. Will this cause problems? @kamilmysliwiec

kamilmysliwiec: Actually, I think you can remove the `Object.defineProperty` call. :)

Diluka: Previously, when dealing with dynamic decorated class generation, if the classname hadn't modified, it won't be working. 😕 

https://github.com/nestjsx/crud/pull/25/files

kamilmysliwiec: Oh, so yeah, there's a risk that someone would use the returned class directly (without extending it by another class). Let's change `Mixin` to `Intersection` then and leave the `Object.defineProperty` call (just in case)

kamilmysliwiec:
Thanks!

PR #823, opened 2020-04-28
Allow specifying `extend type` in a code-first federation approach.

phryneas:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, it is not possible to use the code-first approach to create a federation SDL with the `extend` keyword to reference an entity from an external service.

```diff
-type User @key(fields: "id") @key(fields: "id") {
+extend type User @key(fields: "id") @key(fields: "id") {
      id: ID! @external
      name: String! @external
      nickname: String! @requires(fields: "name")
}
```

## What is the new behavior?

The new service adds the `extendsObjectType` option to the `@ObjectType` annotation.

So the above SDL could be generated by using this code:
```ts

@Directive('@key(fields: "id")')
@ObjectType({ extendsObjectType: true })
export class User {
  @Directive('@external')
  @Field((type) => ID)
  id: string;

  @Directive('@external')
  @Field()
  name: string;

  @Directive('@requires(fields: "name")')
  @Field()
  nickname: string;
}
```

```
[ ] Yes
[X] No
```

## Other information

I'm very open for other suggestions of naming `extendsObjectType` - I just couldn't come up with something more descriptive.

Also, I'll add a docs PR against the main repo after this has been discussed & accepted.

### DIFF
--- a/lib/schema-builder/factories/ast-definition-node.factory.ts
+++ b/lib/schema-builder/factories/ast-definition-node.factory.ts
@@ -9,6 +9,7 @@ import {
   InputValueDefinitionNode,
   ObjectTypeDefinitionNode,
   parse,
+  ObjectTypeExtensionNode,
 } from 'graphql';
 import { head } from 'lodash';
 import { DirectiveParsingError } from '../errors/directive-parsing.error';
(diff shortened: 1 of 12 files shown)


lib/schema-builder/factories/ast-definition-node.factory.ts
```diff
@@ -25,17 +26,32 @@ export class AstDefinitionNodeFactory {
   createObjectTypeNode(
     name: string,
     directiveMetadata?: DirectiveMetadata[],
-  ): ObjectTypeDefinitionNode | undefined {
-    if (isEmpty(directiveMetadata)) {
-      return;
-    }
+  ): ObjectTypeDefinitionNode {
```
phryneas, line 29: The check if a `extend` keyword should be printed by `printSchema` can be found here:

https://github.com/apollographql/apollo-server/blob/e44bb72baafbeaa90885ed7275a7df34c4183c98/packages/apollo-federation/src/service/printFederatedSchema.ts#L191-L192

As you can see, both the AST node and the extension AST node have to be set - so it is no longer possible to do an early bail on this one.

I am pretty sure removing this early bail does not break anything, but someone with more knowledge of the codebase should probably take a second look here.

kamilmysliwiec:
Thank you! I'll review as soon as possible

phryneas:
It seems like this was possible all along, just without documentation - I just noticed there is a different folder with tests for it:

https://github.com/nestjs/graphql/blob/master/tests/code-first-federation/user/user.entity.ts

So I'm gonna close this. Once the frustration sets off I might write some docs for it.

PR #898, opened 2020-05-25
Code first federation unions

### LINKED ISSUE #886: The resolveType option passed to createUnionType is removed when using GraphQLFederationModule (code-first)
**I realize I'm jumping the gun a little here on using the code-first approach with federation, but figured this would be helpful to submit.**

## I'm submitting a...

<pre><code>[x] Bug report</code></pre>

## Current behavior

When using the `GraphQLFederationModule`, `resolveType` is removed from union types created via `createUnionType`, causing the following error in queries that contain that union type:

```
GraphQLError: Abstract type "CartItem" must resolve to an Object type at runtime for field "scan_export.cart_items" with value {...}, received "undefined". 
Either the "CartItem" type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function.
at ensureValidRuntimeType (node_modules/graphql/execution/execute.js:687:11)
at completeAbstractValue (node_modules/graphql/execution/execute.js:680:42)
at completeValue (node_modules/graphql/execution/execute.js:604:12)
at completeValue (node_modules/graphql/execution/execute.js:576:21)
at completeValueCatchingError (node_modules/graphql/execution/execute.js:514:19)
at node_modules/graphql/execution/execute.js:636:25
at from (<anonymous>)
...
```

I dug into `graphql-federation.factory` and it looks like resolveType functions for all UnionTypes are being removed in the `generateSchema` method (comments added by me):

```ts
// Union types in autoGeneratedSchema still have resolveType defined
const autoGeneratedSchema: GraphQLSchema = await this.gqlSchemaBuilder.buildFederatedSchema(
  options.autoSchemaFile,
  options,
  this.resolversExplorerService.getAllCtors()
);
// Union types in executableSchema are all missing resolveType
const executableSchema = buildFederatedSchema({
  typeDefs: gql(printSchema(autoGeneratedSchema)),
  resolvers: this.getResolvers(options.resolvers)
});
```

I'm assuming this is due to the fact that `printSchema` is being used to get the `typeDefs` for `buildFederatedSchema`, but haven't had time to get into potential fixes (though my guess is to do another transform on `executableSchema` where we swap in the Union types from `autoGeneratedSchema`.)

## Expected behavior

The `resolveType` function I provide to `createUnionType` should still be there in the generated schema. I was able to fix this by passing resolvers in the module options, but that shouldn't be required when using `createUnionType`:

```ts
           resolvers: {
                CartItem: {
                    __resolveType(value: CartItemSingle | CartItemGroup) {
                        if (value.type === CartItemTypeEnum.group) {
                            return 'CartItemGroup';
                        }
                        return 'CartItemSingle';
                    }
                }
            }
```

## Minimal reproduction of the problem with instructions

Try to use the `GraphQLFederationModule` with a code-first approach and a union type. All `resolveType` methods on the union types will be missing in the output schema.

## Environment

<pre><code>
"@nestjs/common": "7.0.13",
"@nestjs/core": "7.0.13",
"@nestjs/graphql": "7.3.9",

</code></pre>

parkerself22:
Resolvers for GraphQLUnionType's were being removed when calling buildFederatedSchema
causing queries that returned union's to error

Fixes #886

## PR Checklist

- [ ] (Not applicable) Docs have been added / updated (for bug fixes / features)

## PR Type

```
[x] Bugfix
```

When using the `GraphQLFederationModule` in a code-first approach, queries that return union types created via `createUnionType` error because the apollo-server can't properly resolve the sub-type of the object in the union.

Issue Number: #886

## What is the new behavior?
A `resolveType` method is added to the GraphQLUnionType in the executable schema created via Apollo's `buildFederatedSchema` function.

```
[ ] Yes
[x] No
```

## Other information
@kamilmysliwiec I originally implemented a full fix for this, went to open a PR with it, and saw you had already taken a shot at it by adding `overrideOrExtendResolvers` to `GraphQLFederationFactory`. My PR branch had added E2E test cases, so I merged master, removed my implementation of the fix, and ran my new tests against your fix. Turns out it was still failing, seems to be an underlying issue in `@apollo/federation`. The issue seems to be returning the full `GraphQLObjectType` from the union's resolveType method. It works fine though if you just return the name of the `GraphQLObjectType`. I'd love to know why that's the case but I did a LOT of digging and couldn't figure it out. I do know that returning just the name of the resolved type works in the E2E tests and has been working for us in production (the workaround I mentioned in the issue).


lib/federation/graphql-federation.factory.ts
```diff
@@ -132,9 +135,28 @@ export class GraphQLFederationFactory {
...
+            context,
+            info,
+            abstractType,
+          );
+          if (isString(result)) {
+            return result;
+          }
+          // for some reason this doesnt work if we return the resolved
+          // type object, but does if we return just the name
+          return result.name;
```
parkerself22, line 158: @kamilmysliwiec this is the only difference between the default `resolveType` method created by `ResolveTypeFactory.getResolveTypeFunction`. I originally tried to do it the same way you did but the E2E search query test in `tests/e2e/code-first-federation.spec.ts` still failed with that method.

tests/e2e/code-first.spec.ts
```diff
@@ -43,6 +43,38 @@ describe('Code-first', () => {
       });
   });
 
+  it('should return the search result', () => {
```
parkerself22, line 46: Added this test just to confirm that union types were working properly in non-federated code-first approaches.

parkerself22:
@kamilmysliwiec Just figured out what the problem was, will push a complete solution without the string return hack in a few minutes

parkerself22:
@kamilmysliwiec So I can finally fully explain what the issue with your fix was, and why returning just the name of the resolved type worked. Summary:

- `@apollo/federation` adds additional metadata to all `GraphQLNamedType`'s in the schema under a `federation` property. (see @apollo/federation/src/composition/types.ts:47)
- If you call `resolveType` from the union in the autogenerated schema and get a `GraphQLObjectType`, that object type will be missing the federation property. We return that and `@apollo/federation` will ignore the result.
- Returning just the name of the type worked because apollo will search for a `GraphQLObjectType` in it's federated schema with that name.

I modified how we are overriding the `resolveType` method on the federated union to explicitly find the matching type in the federated schema, figured that would be better for maintenance/readability going forward. My concern now is that there are other issues from overriding federated types that we just haven't seen yet. 

Let me know what you think of this, I could probably find time to keep digging / write tests for the other `GraphQLNamedType`'s that apollo federation adds metadata to.

kamilmysliwiec:
Amazing PR @parkerself22! Thank you for your contribution ❤️

PR #925, opened 2020-06-02
WIP: feat(middleware): field middleware

j:
I quickly put this together for POC.  Borrowed a bit from `type-graphql`.

Container impl isn't correct I don't think.  I just used moduleRef to prototype which caused some tests to fail and I didn't have time to look too much into it.

I was actually expecting my tests where I'm throwing an error in the "AuthMiddleware" to fail fast, but realized (duh, it's been a while), that GraphQL executes promises at the same time, so you should never do anything heavy in any sort of middleware.  Simple context checks would be OK though.  It's how most people are doing it anyway.

### DIFF
--- a/lib/decorators/index.ts
+++ b/lib/decorators/index.ts
@@ -20,3 +20,4 @@ export * from './resolver.decorator';
 export * from './root.decorator';
 export * from './scalar.decorator';
 export * from './subscription.decorator';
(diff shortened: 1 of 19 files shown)


kamilmysliwiec:
Field middleware feature was added in one of the recent releases https://docs.nestjs.com/graphql/field-middleware :)

Issue #993, opened 2020-06-24
Feature: Sort printed autoSchemaFile


mattleff:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

When using the `autoSchemaFile` option the printed schema is not sorted. Additionally, when using forward references between modules the order of the generated schema is indeterminate since the [order of instantiation is indeterminate](https://docs.nestjs.com/fundamentals/circular-dependency#forward-reference). This causes the schema to possibly change every time the server (re)loads.

## Expected behavior

I should be able to provide an option to the `GraphQLModule` to ensure that the printed schema is sorted.

## Minimal reproduction of the problem with instructions

I have attempted to solve this issue with `transformSchema`:

```ts
import { GraphQLSchema, lexicographicSortSchema } from 'graphql';

...

GraphQLModule.forRoot({
    ...
    transformSchema: (schema: GraphQLSchema) => lexicographicSortSchema(schema),
});
```

This correctly and consistently orders the schema for introspection, etc. However `transformSchema` is not applied to the `autoSchemaFile` generation:

https://github.com/nestjs/graphql/blob/e13b51d9f97590b33d8b700ef3816f92a368d8f9/lib/graphql-schema.builder.ts#L74-L81

I am willing to submit a PR to resolve this issue, either by:

- Applying `transformSchema` to the `autoSchemaFile` generation
- Adding a boolean option to `GraphQLModule` to sort the schema (both for introspection, etc. and for `autoSchemaFile`)
- Another direction you suggest

Thanks for Nest and for your guidance here! 👍 

## Environment

<pre><code>
Nest version: 7.0.13
 
For Tooling issues:
- Node version: v14.3.0
- Platform:  Mac

</code></pre>

kamilmysliwiec:
> Applying transformSchema to the autoSchemaFile generation
Adding a boolean option to GraphQLModule to sort the schema (both for introspection, etc. and for autoSchemaFile)

I think both would be very useful! PRs are more than welcome :)

mattleff:
@kamilmysliwiec I've pushed #996 to apply `transformSchema` when writing an `autoSchemaFile`. I can send a PR for a sort option later.

Issue #76, opened 2018-09-25
How to stitch schemas with Nest@5?


merongmerongmerong:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[ ] Feature request
[x] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

```typescript

configure(consumer) {
  const typeDefs = this.graphQLFactory.mergeTypesByPaths('./**/*.graphql');
  const localSchema = this.graphQLFactory.createSchema({ typeDefs });
  const delegates = this.graphQLFactory.createDelegates();
  const schema = mergeSchemas({
    schemas: [localSchema, chirpSchema, linkTypeDefs],
    resolvers: delegates,
  });
  consumer
    .apply(graphqlExpress(req => ({ schema, rootValue: req })))
    .forRoutes('/graphql');
}
```
:shipit: [old doc](https://github.com/nestjs/docs.nestjs.com/blob/master/src/app/homepage/pages/graphql/schema-stitching/schema-stitching.component.ts) 

## Expected behavior

```typescript
async useFactory(graphqlFactory: GraphQLFactory) {
    const link = new HttpLink({ uri: '^_^', fetch });
    const anotherSchema = await introspectSchema(link);
    const remoteSchema = makeRemoteExecutableSchema({ schema: anotherSchema, link });

    return {
      typePaths: ['src/**/*.graphql'],
      schema,
      mergeSchemas(localSchema => ({ 
        schemas: [localSchema, remoteSchema, ...],
        resolvers: graphQLFactory.createDelegates()
      }),
      ...
    };
}
```
:x:  [injecting graphql factory](https://github.com/nestjs/graphql/issues/52) 
:x: accessing local schema 

## What is the motivation / use case for changing the behavior?

[schema stitching](https://www.apollographql.com/docs/graphql-tools/schema-stitching.html) with nest@5

## Environment

<pre><code>
Nest version: 5.3.6

 
For Tooling issues:
- Node version: 10  
- Platform: 
  - OS: Manjaro Linux x86_64 
  - Kernel: 4.14.67-1-MANJARO 

Others:

</code></pre>

patpaquette:
Hi, is there an undocumented way of doing this currently? Wondering if I should dig deeper or wait. Thanks!

idhard:
[...]

PR #996, opened 2020-06-25
feat(): Apply transformSchema to autoSchemaFile

mattleff:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #993 

## What is the new behavior?

The `transformSchema` option is now applied when writing a `autoSchemaFile` to the filesystem.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/lib/graphql-schema.builder.ts
+++ b/lib/graphql-schema.builder.ts
@@ -40,6 +40,7 @@ export class GraphQLSchemaBuilder {
           schemaDirectives: options.schemaDirectives,
         },
         options.sortSchema,
+        options.transformAutoSchemaFile && options.transformSchema,
       );
     } catch (err) {
       if (err && err.details) {
@@ -72,6 +73,7 @@ export class GraphQLSchemaBuilder {
           skipCheck: true,
         },
         options.sortSchema,
+        options.transformAutoSchemaFile && options.transformSchema,
       );
     } catch (err) {
       if (err && err.details) {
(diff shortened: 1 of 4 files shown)


sgarner:
@mattleff This looks like a good change and I love that you are solving the schema sorting problem :+1:

But when applying `transformSchema` to the schema file, I think it might be good to either make this optional, or pass an argument to the `transformSchema` function so that we can customise transformation behavior depending on whether it is being applied for runtime or for file generation.

Some of us are [using `transformSchema` to enable Schema Stitching](https://github.com/nestjs/graphql/issues/76). In that case the result of `transformSchema` will be a gateway schema that combines the local schema with potentially multiple remote schemas. For my use case, I would prefer that the auto schema file only prints the local schema, and not the stitched gateway schema - which will no longer be the case with this PR.

mattleff:
@sgarner Thanks for your feedback! I've made another PR to provide a `sortSchema` option: #997. That PR resolves my issue (#993) so I'll let @kamilmysliwiec decide if he wants this change as well.

kamilmysliwiec:
> But when applying transformSchema to the schema file, I think it might be good to either make this optional, or pass an argument to the transformSchema function so that we can customise transformation behavior depending on whether it is being applied for runtime or for file generation.

To avoid breaking changes, I'd suggest applying `transformSchema` IF a dedicated flag (option) is enabled. :) Would you be able to align PR to such a requirement?

mattleff:
> To avoid breaking changes, I'd suggest applying `transformSchema` IF a dedicated flag (option) is enabled. :) Would you be able to align PR to such a requirement?

Sure. Does `transformAutoFileSchema: boolean` sound right?

kamilmysliwiec:
@mattleff I like it!

mattleff:
@kamilmysliwiec This is ready for review again.

kamilmysliwiec:
Looks great, thank you!

Issue #1001, opened 2020-06-27
Enum as a query variable causes validation errors


dotellie:
## I'm submitting a...

<pre><code>
[x] Regression 
[ ] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

Using an input with an enum type on a query as a variable causes a validation error. A query like this does not work as an example if `UserType` is an enum:

```gql
mutation ($userType: UserType!) {
  addUser(type: $userType) {
    id
    type
  }
}
```

The query would yield the following:

```
"GraphQLError: Variable \"$userType\" of type \"UserType!\" used in position expecting type \"UserType!\"." [...]
```

## Expected behavior

The above query should work without any problems.

## Minimal reproduction of the problem with instructions

Reused my old repro repo again, the query above is from its current state: https://github.com/dotellie/nest-gql-resolve-field-repro

## Regression
The issue started appearing on 7.4.4 and after so it's likely a problem with the custom schema transforms if I had to guess.

dotellie:
After spending a bit time on trying to fix this, I realized it's also an issue for all places where arguments can be provided and it's probably also an issue with input types as well as enums. Basically all types which can be used as inputs are duplicated in the schema. I'm likely going to be able to fix this and do a PR, but I'm probably going the route of copying all of Apollo's `transformSchema` and patching it to not remove additional fields on the objects.

kamilmysliwiec:
Fix published as 7.5.1. :)

Issue #981, opened 2020-06-19
Resolve reference function not being run since 7.3.11


dotellie:
## I'm submitting a...

<pre><code>
[x] Regression 
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

A method on a resolver class with the `ResolveReference` decorator is not run when it should be.

## Expected behavior

The method should run and resolve a reference which can contain all the needed data.

## Minimal reproduction of the problem with instructions

https://github.com/dotellie/nest-gql-resolve-field-repro (I reused an old repro repo)

Clone it, run it, go to localhost:3001/graphql and run the following query
```gql
query($representations: [_Any!]!) {
  _entities(representations: $representations) {
    ... on User {
      id
      somethingElse
    }
  }
}
```
with these variables
```json
{
          "representations": [
            {
              "__typename": "User",
              "id": "hi"
            }
          ]
        }
```

It will fail saying that somethingElse can't be null which it is because the resolve reference was never run. The log statement I have in the resolve reference function also never shows up.

## What is the motivation / use case for changing the behavior?

Federation is basically useless without this. Maybe that isn't so concrete, but I think it's pretty obvious why this needs fixing :sweat_smile: 

## Regression commit
</code></pre>

I managed to nail it down to this commit being the culprit as 7.3.10 still works and this is the only commit in 7.3.11: https://github.com/nestjs/graphql/commit/555c3f48eafdc95604f4319a9d03a00b34d8b0ce

kamilmysliwiec:
Fixed in 7.4.4

PR #1006, opened 2020-06-29
fix(federation): remove duplicate types in schema #1001

dotellie:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1001

## What is the new behavior?
This fix goes back to using `transformSchema` as used before 7.4.4 but because of the bug present in the official Apollo package where federation properties aren't carried over, this PR has that entire file copied over locally with a somewhat hacky fix. Unless Apollo suddenly changes federation however, the fix should be stable. The best solution is of course that Apollo patches their package and we can revert to using that, but I think for now this change is good to have, just so all the problems that have been popping up (such as #1001 and #981) can be fixed.

```
[ ] Yes
[x] No
```

## Other information

I added some tests for these cases so they don't break again :blush:


dotellie:
This PR took a while to get up because of the GitHub outage today, but here we are finally. I just noticed that I forgot to remove unused imports in the factory file (I'm used to having eslint complaining at me for that :sweat_smile:) but I don't exactly know how I should clear that up. I'm thinking of squashing both the current commits and doing that change at the same time. Thoughts?

dotellie:
@kamilmysliwiec Any chance you could take a look at this soon?

lib/utils/transform-schema.util.ts
```diff
@@ -0,0 +1,169 @@
+// This file is copied from `apollo-tooling`. The only difference is that it has a hack to not remove federation specific properties.
+// https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo-graphql/src/schema/transformSchema.ts
```
kamilmysliwiec, line 2: Can we add a note of what lines specifically must have been changed?

lib/federation/graphql-federation.factory.ts
```diff
@@ -199,18 +201,10 @@ export class GraphQLFederationFactory {
...
-      } else if (isScalarType(type) && type.name === 'DateTime') {
-        const autoGeneratedScalar = autoGeneratedSchema.getTypeMap()[
-          key
-        ] as GraphQLScalarType;
-        if (!autoGeneratedScalar) {
-          return;
-        }
-        type.parseLiteral = autoGeneratedScalar.parseLiteral;
-        type.parseValue = autoGeneratedScalar.parseValue;
+        return type;
```
kamilmysliwiec, line 204: What about scalars?

kamilmysliwiec:
Thanks @dotellie! I've left 2 minor comments

dotellie:
@kamilmysliwiec I think both your comments should be fixed now although I'm not completely sure about the scalar problem since I didn't understand it to begin with and there are no test cases for it.

kamilmysliwiec:
Thank you @dotellie!

> I'm not completely sure about the scalar problem since I didn't understand it to begin with and there are no test cases for it

It basically inherits the `DateTime` scalar from the automatically generated schema :)

PR #1027, opened 2020-07-09
feat(decorators): add gql-headers decorator

Edgar-P-yan:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
One of my projects requires direct access to the request headers. I implemented this decorator in my code, and thought maybe it is worth to add to the package.
In summary, I added `GqlHeaders` parameter decorator, generally the analogue of `Headers` decorator in `@nestjs/common`.

Example usage:
```ts
import { IncomingHttpHeaders } from 'http';

@Resolver()
export class GraphqlApiResolver {
  @Query(() => String)
  content(
    @GqlHeaders('User-Agent') userAgent: string,
    @GqlHeaders() allHeaders: IncomingHttpHeaders,
  ): Promise<string> {
    return service.getSpecialContent({ userAgent });
  }
}
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/lib/decorators/gql-headers.decorator.ts
+++ b/lib/decorators/gql-headers.decorator.ts
@@ -1,0 +1,29 @@
+import { createParamDecorator, ExecutionContext } from '@nestjs/common';
+import { IncomingMessage } from 'http';
+import { GqlExecutionContext } from '../services';
+
+/**
+ * `@GqlHeaders()` Route handler parameter decorator. Extracts the `headers`
+ * property from the `req` object and populates the decorated
+ * parameter with the value of `headers`.
+ *
+ * For example: `async update(@GqlHeaders('Cache-Control') cacheControl: string)`
+ *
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
We had several similar discussions around this topic. I believe that adding more and more decorators that simply pick properties from the request instance might be redundant. The request object itself has so many properties that can be important to certain applications that we could theoretically create tons of additional decorators, but taking into account how simple it is (+ it's thoroughly described in the docs), it probably doesn't make sense. Thank you for your contribution though!

Edgar-P-yan:
@kamilmysliwiec Hi! i got your point. Actually after this PR while i was using this and several other custom decorators i noticed that having decorator for each little feature is not a very good idea.
Thanks!

Issue #1248, opened 2020-11-18
[discussion] Specify TypeScript types for custom scalars


david-golightly-leapyear:
Feature Request.

I'm having the same issue as https://github.com/nestjs/graphql/issues/540, but I'd like to take my own stab at explaining it, and why the resolution of this issue is unsatisfactory.

I'm a big fan of TypeScript and having consistently applied, strong types throughout my server code. To support that, I'm using https://github.com/urigo/graphql-scalars for more specialized scalars, to support custom scalar types like BigInt, DateTime etc.

I've defined application-custom scalars in `base.graphql`:

```graphql
"""
A date-time string at UTC, such as 2019-12-03T09:54:33Z, compliant with the date-time format.
"""
scalar DateTime

"""
An arbitrary JSON blob
"""
scalar JSON
```

And have supplied appropriate resolvers for them in the GraphQL module, etc. This all works fine, and GraphQL correctly serializes and deserializes the scalar values.

However, the generated TypeScript types are insufficient:

```ts
export type DateTime = any;
export type JSON = any;
```

Even though I've defined my DateTime resolver to parse to/serialize from a JS `Date` object, I'm unable to leverage that knowledge - that my application is in full control of - to know that a GraphQL argument of type `DateTime` is actually a `Date`. Instead, I have a bunch of `any` values. And in my application, I go to pains to avoid the use of `any`.

So first of all, a more appropriate default "unknown" type would be TypeScript 3.0's ['unknown`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type). This would force explicit (rather than implicit) casting, and require assumptions to be made clear.

But secondly, like the author of the original ticket, I've used [`graphql-codegen`](https://graphql-code-generator.com/) and think its model works fairly well as an example of what might be possible in a framework as ambitious as Nest.js.

In graphql-codegen, I can manually specify how I want my custom GraphQL scalar values to be mapped to TypeScript like so:

```yml
config:
  scalars:
    DateTime: 'Date'
    JSON: 'JSON'
```

Now I'm able to generate the following TypeScript:

```ts
export type Scalars = {
  DateTime: Date
  JSON: JSON
}
```

This is the functionality I'm looking for.

Of course, if you want to get technical about it, Nest.js has some shortcomings in terms of its interaction with TypeScript and build-time type safety. For example, as an argument to a resolver query or mutation, I can write:

```ts
@Args('date') date: Date
```

[...]

PR #1253, opened 2020-11-21
Specify generated types for custom scalars

davidgoli:
## PR Checklist

I'd be happy to contribute documentation, but it wasn't obvious to me where to find its source.

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/nestjs/graphql/issues/1248

## What is the new behavior?

Adds the following 3 fields to `DefinitionsGeneratorOptions`:

```ts

  /**
   * If provided, specifies a default generated TypeScript type for custom scalars.
   * @default 'any'
   */
  defaultScalarType?: string;

  /**
   * If provided, specifies a mapping of types to use for custom scalars
   * @default undefined
   */
  customScalarTypeMapping?: Record<string, string>;

  /**
   * If provided, specifies a custom header to add to the output file (eg. for custom type imports)
   * @default undefined
   */
  header?: string;
```

Sample usage:

```graphql
scalar List
scalar Point
scalar BigNumber
```

```ts
    await graphqlFactory.generateDefinitions(typeDefs, {
      definitions: {
        path: outputFile,
        defaultScalarType: 'unknown',
        customScalarTypeMapping: {
          List: 'string[]',
          Point: '[number, number]',
          BigNumber: 'BigNumber',
        },
        header: "import { BigNumber as _BigNumber } from 'bignumber.js'",
      },
    });
```

emits:

```ts

/** ------------------------------------------------------
 * THIS FILE WAS AUTOMATICALLY GENERATED (DO NOT MODIFY)
 * -------------------------------------------------------
 */

/* tslint:disable */
/* eslint-disable */
import { BigNumber as _BigNumber } from 'bignumber.js'

export type List = string[];
export type Point = [number, number];
export type BigNumber = _BigNumber;
```

```
[ ] Yes
[x] No
```

The new properties are optional and do not break backwards compatibility.

## Other information
There are some additional diffs since I ran `npm run format` as suggested in the [coding rules](https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md#-coding-rules)

### DIFF
--- a/lib/graphql-ast.explorer.ts
+++ b/lib/graphql-ast.explorer.ts
@@ -46,6 +46,25 @@ export interface DefinitionsGeneratorOptions {
    * @default false
    */
   skipResolverArgs?: boolean;
+
+  /**
+   * If provided, specifies a default generated TypeScript type for custom scalars.
+   * @default 'any'
+   */
+  defaultScalarType?: string;
+
+  /**
+   * If provided, specifies a mapping of types to use for custom scalars
+   * @default undefined
+   */
+  customScalarTypeMapping?: Record<string, string | { name: string }>;
+
+  /**
+   * If provided, specifies a custom header to add after the
+   * to the output file (eg. for custom type imports or comments)
+   * @default undefined
+   */
+  additionalHeader?: string;
 }
 
 @Injectable()
@@ -86,7 +105,10 @@ export class GraphQLAstExplorer {
       ),
     );
 
-    tsFile.insertText(0, DEFINITIONS_FILE_HEADER);
+    const header = options.additionalHeader
+      ? `${DEFINITIONS_FILE_HEADER}\n${options.additionalHeader}\n\n`
+      : DEFINITIONS_FILE_HEADER;
+    tsFile.insertText(0, header);
     return tsFile;
   }
 
@@ -113,7 +135,7 @@ export class GraphQLAstExplorer {
         return this.addObjectTypeDefinition(item, tsFile, 'interface', options);
       case 'ScalarTypeDefinition':
       case 'ScalarTypeExtension':
-        return this.addScalarDefinition(item, tsFile);
+        return this.addScalarDefinition(item, tsFile, options);
       case 'EnumTypeDefinition':
       case 'EnumTypeExtension':
         return this.addEnumDefinition(item, tsFile);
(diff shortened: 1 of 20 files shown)


lib/graphql-ast.explorer.ts
```diff
@@ -46,6 +46,24 @@ export interface DefinitionsGeneratorOptions {
...
+   * If provided, specifies a mapping of types to use for custom scalars
+   * @default undefined
+   */
+  customScalarTypeMapping?: Record<string, string>;
+
+  /**
+   * If provided, specifies a custom header to add to the output file (eg. for custom type imports)
+   * @default undefined
+   */
+  header?: string;
```
kamilmysliwiec, line 66: The `header` name may be somewhat ambigious. Maybe we should use `importStatements` instead(?)

kamilmysliwiec:
The PR looks great overall! 🎉 I left one minor comment

lib/graphql-ast.explorer.ts
davidgoli: Good feedback! I might avoid using the term "import" because, while that's what I'm using it for here, it's really just an open-ended string that gets appended to the default header, and before the rest of the document - could be comments or other content, not necessarily an import statement. I'm open to suggestions!

kamilmysliwiec: Thanks for the clarification! In this case, we might need to emphasize somehow that this header will not **override** the existing one but will be appended to the default one. With the "header" configuration property name, I'd rather think (as a developer) that I can override the default one. Any ideas?

davidgoli: `addtionalHeader` or `subHeader` perhaps?

mattleff: I like `additionalHeader`, FWIW.

davidgoli: Done

kamilmysliwiec:
I think we should be ready to merge this PR. Would you be able to create a PR to the docs with this feature (this chapter https://docs.nestjs.com/graphql/scalars)?

davidgoli:
@kamilmysliwiec Documentation updated with https://github.com/nestjs/docs.nestjs.com/pull/1573

lib/graphql-ast.explorer.ts
```diff
@@ -46,6 +46,25 @@ export interface DefinitionsGeneratorOptions {
...
+   * If provided, specifies a default generated TypeScript type for custom scalars.
+   * @default 'any'
+   */
+  defaultScalarType?: string;
+
+  /**
+   * If provided, specifies a mapping of types to use for custom scalars
+   * @default undefined
+   */
+  customScalarTypeMapping?: Record<string, string>;
```
kamilmysliwiec, line 60: @davidgoli One last minor improvement (should be easy to add) - what do you think about supporting the `Record<string, string | Function>` type instead? Example usage:

```typescript
customScalarTypeMapping: {
  DateTime: Date,
},
```
Internally, we would simply check if a value is === 'function' and if so, read the `.name` property (e.g., `Date.name`). The rest would remain the same

davidgoli: Done, and pushed!

kamilmysliwiec:
Thanks for this amazing contribution!

PR #1292, opened 2020-12-17
feat(@nestjs/graphql): base exception filter for apollo exceptions

### LINKED ISSUE #1053: GraphQL error handling
Nest version: lastest.

Throwing errors with Nest HttpException does not play well with GraphQL.

`throw new UnauthorizedException()` will generate the following:

```
{
  "errors": [
    {
      "message": "Unauthorized",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "login"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "response": {
            "statusCode": 401,
            "message": "Unauthorized"
          },
          "status": 401,
          "message": "Unauthorized",
          "stacktrace": [stack]
        }
      }
    }
  ],
  "data": null
}
```

While `throw new AuthenticationError("AuthenticationError")` from `import { AuthenticationError } from "apollo-server-express"` will generate the proper friendly error:

```
{
  "errors": [
    {
      "message": "AuthenticationError",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "login"
      ],
      "extensions": {
        "code": "UNAUTHENTICATED",
        "exception": {
          "stacktrace": [stack]
        }
      }
    }
  ],
  "data": null
}
```

When using the GraphQL the error responses should be in second format.

https://www.apollographql.com/docs/apollo-server/data/errors/

nirga:
Closes #1053

## PR Checklist

- [X] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Exceptions are built by Apollo server engine, and because it doesn't recognizes Nest.js's exceptions, it always returns a generic HTTP 500 error, even for other errors, specifically UNAUTHORIZED. 

Issue Number: #1053 

## What is the new behavior?
Exceptions are converted explicitly to an `ApolloError`, possibly to a specific exception (if exists), and then thrown for the Apollo server to catch and format. 

```
[X] Yes
[ ] No
```

This change might cause breakages if clients used to assume HTTP 500 is always returned. However, this will now be in-line with the rest of Apollo server errors and exceptions so it should work well with compliant clients.

## Other information

### DIFF
--- a/lib/graphql-base-exception-filter.ts
+++ b/lib/graphql-base-exception-filter.ts
@@ -1,0 +1,39 @@
+import { Catch, ArgumentsHost, HttpException } from '@nestjs/common';
+import { BaseExceptionFilter } from '@nestjs/core';
+import {
+  ApolloError,
+  AuthenticationError,
+  ForbiddenError,
+} from 'apollo-server-errors';
+import { GqlContextType } from './services';
+
+const apolloPredefinedExceptions: Record<number, typeof ApolloError> = {
+  401: AuthenticationError,
+  403: ForbiddenError,
+};
+
+@Catch(HttpException)
+export class GraphQLBaseExceptionFilter extends BaseExceptionFilter {
+  catch(exception: HttpException, host: ArgumentsHost) {
+    if (host.getType<GqlContextType>() !== 'graphql') {
+      return null;
+    }
+
+    let error: ApolloError;
+    if (exception.getStatus() in apolloPredefinedExceptions) {
+      error = new apolloPredefinedExceptions[exception.getStatus()](
+        exception.message,
+      );
+    } else {
+      error = new ApolloError(
+        exception.message,
+        exception.getStatus().toString(),
+      );
+    }
+
+    error.stack = exception.stack;
+    error.extensions['response'] = exception.getResponse();
+
+    throw error;
+  }
+}
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -2,6 +2,7 @@ export * from './decorators';
 export * from './federation';
 export * from './graphql-ast.explorer';
 export * from './graphql-definitions.factory';
+export * from './graphql-base-exception-filter';
 export * from './graphql-schema.host';
 export * from './graphql-types.loader';
 export * from './graphql.factory';
--- a/tests/e2e/guards-filters.spec.ts
+++ b/tests/e2e/guards-filters.spec.ts
@@ -1,6 +1,7 @@
 import { INestApplication } from '@nestjs/common';
 import { Test } from '@nestjs/testing';
 import * as request from 'supertest';
+import { GraphQLBaseExceptionFilter } from '../../lib';
 import { ApplicationModule } from '../code-first/app.module';
 
 describe('GraphQL - Guards', () => {
@@ -12,6 +13,7 @@ describe('GraphQL - Guards', () => {
     }).compile();
 
     app = module.createNestApplication();
+    app.useGlobalFilters(new GraphQLBaseExceptionFilter());
     await app.init();
   });
 
@@ -26,7 +28,7 @@ describe('GraphQL - Guards', () => {
       .expect(200, {
         errors: [
           {
-            message: 'Unauthorized error',
+            message: 'Unauthorized',
             locations: [
               {
                 line: 2,
@@ -35,7 +37,11 @@ describe('GraphQL - Guards', () => {
             ],
             path: ['recipe'],
             extensions: {
-              code: 'INTERNAL_SERVER_ERROR',
+              code: 'UNAUTHENTICATED',
+              response: {
+                statusCode: 401,
+                message: 'Unauthorized',
+              },
             },
           },
         ],
--- a/tests/e2e/pipes.spec.ts
+++ b/tests/e2e/pipes.spec.ts
@@ -1,6 +1,7 @@
 import { INestApplication, ValidationPipe } from '@nestjs/common';
 import { Test } from '@nestjs/testing';
 import * as request from 'supertest';
+import { GraphQLBaseExceptionFilter } from '../../lib';
 import { ApplicationModule } from '../code-first/app.module';
 
 describe('GraphQL - Pipes', () => {
@@ -12,6 +13,7 @@ describe('GraphQL - Pipes', () => {
     }).compile();
 
     app = module.createNestApplication();
+    app.useGlobalFilters(new GraphQLBaseExceptionFilter());
     app.useGlobalPipes(new ValidationPipe());
     await app.init();
   });
@@ -30,17 +32,13 @@ describe('GraphQL - Pipes', () => {
         errors: [
           {
             extensions: {
-              code: 'INTERNAL_SERVER_ERROR',
-              exception: {
-                message: 'Bad Request Exception',
-                response: {
-                  error: 'Bad Request',
-                  message: [
-                    'description must be longer than or equal to 30 characters',
-                  ],
-                  statusCode: 400,
-                },
-                status: 400,
+              code: '400',
+              response: {
+                error: 'Bad Request',
+                message: [
+                  'description must be longer than or equal to 30 characters',
+                ],
+                statusCode: 400,
               },
             },
             locations: [


lib/graphql.module.ts
```diff
@@ -46,6 +47,7 @@ import {
     GraphQLTypesLoader,
     GraphQLSchemaBuilder,
     GraphQLSchemaHost,
+    { provide: APP_FILTER, useClass: GraphQLExceptionFilter },
```
jmcdo29, line 50: This will bind the filter for the entire application. If someone is using GQL and REST they are now using the `GraphQLExceptionFilter` for both contexts

nirga: I've added a check in `GraphQLExceptionFilter` to rethrow exceptions that were thrown not in GraphQL context.
If you think it make sense, I can go ahead and add a test for that specifically. I'd probably need to add a REST controller to the *code-first* test app, does it make sense?

kamilmysliwiec: To avoid introducing breaking changes, let's make this optional for now and inform users that they have to explicitly register this filter in their applications. Later (in v8), we can get back to this discussion and think about registering it automatically.

lib/graphql-exception.filter.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const apolloPredefinedExceptions: Record<number, typeof ApolloError> = {
+  401: AuthenticationError,
+  403: ForbiddenError,
+};
+
+@Catch(HttpException)
+export class GraphQLExceptionFilter implements ExceptionFilter {
+  catch(exception: HttpException, host: ArgumentsHost) {
+    if (host.getType<GqlContextType>() != 'graphql') {
```
kamilmysliwiec, line 22:
```suggestion
    if (host.getType<GqlContextType>() !== 'graphql') {
```

lib/graphql-exception.filter.ts
```diff
@@ -0,0 +1,43 @@
...
+  Catch,
+  ArgumentsHost,
+  HttpException,
+  ExceptionFilter,
+} from '@nestjs/common';
+import {
+  ApolloError,
+  AuthenticationError,
+  ForbiddenError,
+} from 'apollo-server-express';
```
kamilmysliwiec, line 11: We can't import from the 'apollo-server-express' as this will break Fastify based applications. Can you import from base/core packages (not sure which one exposes these error classes)

kamilmysliwiec:
Thanks for your contribution! I've left a few comments/suggestions

lib/graphql.module.ts
nirga: Sure. I've added it manually to *guards-filters.spec.ts* and *pipes.spec.ts* as these are the tests that throw exceptions.

lib/apollo-exception.filter.ts
```diff
@@ -0,0 +1,39 @@
...
+const apolloPredefinedExceptions: Record<number, typeof ApolloError> = {
+  401: AuthenticationError,
+  403: ForbiddenError,
+};
+
+@Catch(HttpException)
+export class ApolloExceptionFilter implements GqlExceptionFilter {
+  catch(exception: HttpException, host: ArgumentsHost) {
+    if (host.getType<GqlContextType>() !== 'graphql') {
+      throw exception;
```
faboulaws, line 19: Is it ok to throw an exception here? From what I understand it not ok

> The ExceptionFilter is always the last place that gets called before a response is sent out, it is responsible for building the response. You cannot rethrow an exception from within an ExceptionFilter.

Since this PR was not merged I copied this class to our own project. Our project has both a graphQL and REST routes. This line is crashing the App. I had to make the following change to make it work 

```javascript
 if (host.getType<GqlContextType>() !== 'graphql') {
      const ctx = host.switchToHttp();
      const response = ctx.getResponse<Response>();
      const status = exception.getStatus();

      response
        .status(status)
        .json({
          statusCode: status,
          message: exception.message
        });
      return exception;
    }
```

I am still new to NestJS. SO am I missing something here?

lib/apollo-exception.filter.ts
```diff
@@ -0,0 +1,39 @@
+import { Catch, ArgumentsHost, HttpException } from '@nestjs/common';
```
faboulaws, line 1: I wondering if this class is needed. This feels more like an example on how to implement exception filters for GraphQL. Most people will not use this directly  since custom logic such as logging etc might be required depending on use case.  Maybe as a base class similar to `BaseExceptionFilter`.

Another approach is to define a helper to convert HTTPExceptions to Apollo Errors.

```javascript
import { ApolloError, AuthenticationError, ForbiddenError } from 'apollo-server-errors'
import { HttpException } from '@nestjs/common'

const apolloPredefinedExceptions: Record<number, typeof ApolloError> = {
  401: AuthenticationError,
  403: ForbiddenError
}

export function convertErrorToApolloError(exception: HttpException): ApolloError {
  let error: ApolloError
  if (exception.getStatus() in apolloPredefinedExceptions) {
    error = new apolloPredefinedExceptions[exception.getStatus()](
      exception.message
    )
  } else {
    error = new ApolloError(
      exception.message,
      exception.getStatus().toString()
    )
  }

  error.stack = exception.stack
  error.extensions['response'] = exception.getResponse()
  return error
}
``` 

Then this helper can be used in any custom `ExceptionFilter` implementing `GqlExceptionFilter` like below.

```javascript
@Catch(HttpException)
export class MyCustomExceptionFilter implements GqlExceptionFilter {
  catch(exception: HttpException, host: ArgumentsHost) {
   const apolloError: ApolloError = convertErrorToApolloError(exception)

    this.logger.error(apolloError)
    return apolloError;
  }
}
```

lib/apollo-exception.filter.ts
```diff
@@ -0,0 +1,39 @@
...
+      error = new ApolloError(
+        exception.message,
+        exception.getStatus().toString(),
+      );
+    }
+
+    error.stack = exception.stack;
+    error.extensions['response'] = exception.getResponse();
+
+    throw error;
```
faboulaws, line 37: Based on the `GqlExceptionFilter ` interface you should return here.

lib/apollo-exception.filter.ts
nirga: You're right, thanks for flagging! I've changed it though I preferred to inherit from `BaseExceptionFilter` and just call its `.catch` to keep the logic consistent with the rest of Nest.

lib/apollo-exception.filter.ts
nirga: I think this class should be as part of the infra so you get this behavior "out of the box" without needing to worry about the internals of Nest exceptions vs. Apollo exceptions. And this is indeed the plan (see @kamilmysliwiec comment above), but I'm changing its name to be `BaseGqlExceptionFilter` for now so it'll be clear that you should inherit from it in GraphQL environments.

lib/apollo-exception.filter.ts
nirga: That's on purpose though. I want ApolloServer to catch this exception so it can format it nicely in the response.

lib/graphql-base-exception-filter.ts
```diff
@@ -0,0 +1,39 @@
+import { Catch, ArgumentsHost, HttpException } from '@nestjs/common';
+import { BaseExceptionFilter } from '@nestjs/core';
+import {
+  ApolloError,
+  AuthenticationError,
+  ForbiddenError,
+} from 'apollo-server-errors';
+import { GqlContextType } from './services';
+
+const apolloPredefinedExceptions: Record<number, typeof ApolloError> = {
```
faboulaws, line 10: Any reason why `UserInputError` and other ApolloErrors are not included?

incompletude:
Guys, it would be nice to have this merged ASAP. Lots of if graphql if rest in my code.

incompletude:
Is there a timeline for this PR? I need it really bad.

kimroen:
> To avoid introducing breaking changes, let's make this optional for now and inform users that they have to explicitly register this filter in their applications. Later (in v8), we can get back to this discussion and think about registering it automatically.

@kamilmysliwiec Based on this comment it looks like this was slated for a 7.x release, but due to the recent base change and merge into the `nestjs:8.0.0` branch, I'm assuming this is no longer the case.

Would that be accurate? Just want to make sure I'm not just missing this landing in a 7.x release, I'm totally fine with doing workarounds until 8 🙂

kamilmysliwiec:
Correct @kimroen

viztor:
> Correct @kimroen

May I suggest v7 to include GraphQLBaseExceptionFilter for developers to register themselves.  That could smooth the upgrade to v8 later. It is suggested in the comment that it would be done like that but it seems it's not there yet.

ScreamZ:
Had issue tracking where this code belongs to in v8 now, in order to understand its behavior, for those who get interested in like me just check : https://github.com/nestjs/graphql/blob/83b4919a45ce459da31e9c40813ef1f00cd5a43d/lib/utils/merge-defaults.util.ts#L78

abouroubi:
Looks like this was removed from version 9. 
Is there a new way to handle exception in graphql now ?

JVMartin:
I'm confused... in v8, do I need to register a GraphQLExceptionFilter like this?

```
app.useGlobalFilters(GraphQLExceptionFilter);
```

I can't find this exported anywhere, so I'm not sure how to set this up. Trying to throw Apollo exceptions in v8 graphql, and they are logged as errors when they shouldn't be.

Distortedlogic:
Can we get some docs on the built-in nestjs patterns for handling graphql expection filtering? When I google, I dont get much thats informative or get things that are old. And I cannot find anything on doing it for subscriptions.. Trial and error to figure this out isnt preferred and means I potentially miss nestjs stuff that would simplify things.

dijonkitchen:
This might have moved to here: https://github.com/nestjs/graphql/blob/master/packages/graphql/lib/interfaces/gql-exception-filter.interface.ts

PR #1348, opened 2021-01-26
feat: allow custom gql-on-ws protocol for subscriptions

### LINKED ISSUE #1341: [discussion]: Custom GraphQL over WebSocket protocol when enabling subscriptions
When [enabling subscriptions](https://github.com/nestjs/graphql/blob/master/lib/graphql.module.ts#L156), the GraphQL module relies on the [default behaviour](https://github.com/apollographql/apollo-server/blob/main/packages/apollo-server-core/src/ApolloServer.ts#L599) of Apollo Server, which is to use [`subscriptions-transport-ws`](https://github.com/apollographql/subscriptions-transport-ws). The library is not actively maintained anymore, and their README points to `graphql-ws` as an alternative. Apollo Server has still `subscription-transport-ws` in their docs (see [here](https://www.apollographql.com/docs/react/data/subscriptions/#setting-up-the-transport)), but according to their [roadmap](https://github.com/apollographql/apollo-server/blob/main/ROADMAP.md#tech-debt-paid-down) they want to remove it somewhere in the future.

`subscriptions-transport-ws` has several protocol bugs that haven't been addressed in the last year. One of them is [this one](https://github.com/apollographql/subscriptions-transport-ws/issues/349), which allows connections to be established while skipping the recommended way to perform [authentication](https://www.apollographql.com/docs/graphql-subscriptions/authentication/). Other protocol bugs are on the client side, and they [keep coming](https://github.com/apollographql/subscriptions-transport-ws/issues/808) even after the [last batch](https://github.com/apollographql/subscriptions-transport-ws/issues/777#issuecomment-674835070) of bug fixes, and [old bugs](https://github.com/apollographql/subscriptions-transport-ws/issues/339) have still not been addressed.

`subscriptions-transport-ws` is clearly not production ready. I've seen the impact of all of these bugs on the project I'm currently working on, and fixed some of them by forking the library and using the fork in our NestJS subscription servers, but we are desperate for a long term solution. Switching from `subscriptions-transport-ws` to `graphql-ws` (as suggested on the first library repo) cannot be done transparently: the two protocols have differences in the structure of messages that are exchanged between client and server that make them incompatible, so which protocol to use should be left as a choice to the user.

What I'd like is for the GraphQL module to allow the user to choose which GraphQL over WebSocket protocol to use, while still defaulting to `subscriptions-transport-ws`. It should be possible, since `subscriptions-transport-ws` is not directly referenced anywhere in this repo, but I still haven't made an attempt. I will try as soon as I can and update this issue.

**EDIT:** I've put together a simple and rough integration of `graphql-ws`. It's available [here](https://github.com/asmeikal/graphql/tree/graphql-ws) and it's in no way good, but it works. The new protocol is enabled by passing `protocol: 'graphql-ws'` to the `subscriptions` option of the GraphQL module. The `subscriptions` option is a union to allow this (see [here](https://github.com/asmeikal/graphql/blob/graphql-ws/lib/interfaces/gql-module-options.interface.ts#L35)). Using the `graphql-ws` protocol is relatively straight forward (see [here](https://github.com/asmeikal/graphql/blob/graphql-ws/lib/graphql.module.ts#L160)), but that code is relying on [this function](https://github.com/enisdenjo/graphql-ws/blob/master/src/use/ws.ts#L30) provided by `graphql-ws`, and the authors say that it's just an example implementation.

I've added a [test](https://github.com/asmeikal/graphql/blob/graphql-ws/tests/subscriptions/subscription-transport-ws.spec.ts) to make sure that the `subscriptions-transport-ws` protocol still works, and [one](https://github.com/asmeikal/graphql/blob/graphql-ws/tests/subscriptions/graphql-ws.spec.ts) to test the `graphql-ws` protocol.

asmeikal:
This could close #1341 for the moment, and provide a way to use `graphql-ws` instead of `subscriptions-transport-ws` without breaking existing NestJS services that already use subscriptions.

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1341 

As described in the issue, the GraphQL module currently relies on `subscriptions-transport-ws`, which has several issues and is being deprecated in favor of `graphql-ws`. Apollo will change the gql-on-ws protocol in the future, but I don't believe `subscriptions-transport-ws` to be production ready, and cannot wait for Apollo to switch protocols nor renounce to NestJS.

## What is the new behavior?

It's now possible to specify a `protocol` when passing `subscriptions` options to the GraphQL module.
The new protocol is `graphql-ws`, and has a few differences in the signature of the callback that the user can provide.
The signatures match the ones documented in `graphql-ws`.

```
[ ] Yes
[x] No
```

## Other information

The `GraphqlWsSubscriptionService` is taken from the [example implementation](https://github.com/enisdenjo/graphql-ws/blob/master/src/use/ws.ts) provided by `graphql-ws`. The purpose of this class is to connect the WebSocket server, the `graphql-ws` server, and Apollo. The main difference with the example implementation is in how errors are handled.

The `GraphqlWsException` class can be used to close the WebSocket connection with a specific code and reason, as suggested in the `graphql-ws` documentation to implement custom authentication protocols (see [here](https://github.com/enisdenjo/graphql-ws#ws-auth-handling)). The `onConnect` callback can use this exception to signal errors to the client, and the client may behave differently based on the error, e.g., it may refresh the authentication token if it is expired. The tests cover this part.

The `graphql-ws` server is highly customizable via callbacks, that are called by the protocol in different phases of the subscription. Most of these are quite advanced and "low level", and allow manipulation of the `ExecutionArgs` or of the protocol message, so I only included the `onConnect`, `onDisconnect`, and `onClose` ones in the GraphQL module `subscriptions` configuration.

A catch: while the new library is called `graphql-ws`, its [WebSocket sub-protocol](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#communication) is called `graphql-transport-ws`, and the [WebSocket sub-protocol](https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/protocol.ts#L1) of `subscriptions-transport-ws` is called `graphql-ws`! :disappointed:  Maybe the `protocol` option passed to the `subscriptions` configuration should be named after the actual WebSocket sub-protocol, and not after the library, to avoid confusion.

### DIFF
--- a/lib/graphql-ws/graphql-ws-subscription.service.ts
+++ b/lib/graphql-ws/graphql-ws-subscription.service.ts
@@ -1,0 +1,135 @@
+import { makeServer, ServerOptions } from 'graphql-ws';
+import * as ws from 'ws';
+import { IncomingMessage } from 'http';
+import { execute, subscribe } from 'graphql';
+import { GraphqlWsException } from './graphql-ws.exception';
+
+export type WebSocket = typeof ws.prototype;
+
+export interface GraphqlWsSubscriptionServiceOptions {
+  schema: ServerOptions['schema'];
+  onConnect?: ServerOptions['onConnect'];
+  onDisconnect?: ServerOptions['onDisconnect'];
+  context?: ServerOptions['context'];
+  keepAlive?: number;
+}
+
+export interface Extra {
+  readonly socket: WebSocket;
+  readonly request: IncomingMessage;
+}
+
+export class GraphqlWsSubscriptionService {
+  private readonly wss: ws.Server;
+
+  constructor(
+    private readonly options: GraphqlWsSubscriptionServiceOptions,
+    server: any,
+  ) {
+    this.wss = new ws.Server({
+      server,
+      path: '/graphql',
+    });
+    this.initialize();
+  }
+
+  private initialize() {
+    const server = makeServer<Extra>({
+      schema: this.options.schema,
+      execute,
+      subscribe,
+      context: this.options.context,
+      onConnect: this.options.onConnect,
+      onDisconnect: this.options.onDisconnect,
+    });
+
+    const ws = this.wss;
+
+    ws.on('error', (err) => {
+      // catch the first thrown error and re-throw it once all clients have been notified
+      let firstErr: Error | null = err;
+
+      // report server errors by erroring out all clients with the same error
+      for (const client of ws.clients) {
+        try {
+          client.close(1011, 'Internal Error');
+        } catch (err) {
+          firstErr = firstErr ?? err;
+        }
+      }
+
+      if (firstErr) {
+        throw firstErr;
+      }
+    });
+
+    const keepAlive = this.options.keepAlive;
+
+    ws.on('connection', (socket, request) => {
+      // keep alive through ping-pong messages
+      let pongWait: NodeJS.Timeout | null = null;
+      const pingInterval =
+        keepAlive > 0 && isFinite(keepAlive)
+          ? setInterval(() => {
+              // ping pong on open sockets only
+              if (socket.readyState === socket.OPEN) {
+                // terminate the connection after pong wait has passed because the client is idle
+                pongWait = setTimeout(() => {
+                  socket.terminate();
+                }, keepAlive);
+
+                // listen for client's pong and stop socket termination
+                socket.once('pong', () => {
+                  if (pongWait) {
+                    clearTimeout(pongWait);
+                    pongWait = null;
+                  }
+                });
+
+                socket.ping();
+              }
+            }, keepAlive)
+          : null;
+
+      const closed = server.opened(
+        {
+          protocol: socket.protocol,
+          send: (data) =>
+            new Promise((resolve, reject) => {
+              socket.send(data, (err) => (err ? reject(err) : resolve()));
+            }),
+          close: (code, reason) => socket.close(code, reason),
+          onMessage: (cb) =>
+            socket.on('message', async (event) => {
+              try {
+                await cb(event.toString());
+              } catch (err) {
+                if (err instanceof GraphqlWsException) {
+                  socket.close(err.code, err.reason);
+                } else {
+                  socket.close(1011, 'Internal error');
+                }
+              }
+            }),
+        },
+        { socket, request },
+      );
+
+      socket.once('close', (code, reason) => {
+        if (pongWait) clearTimeout(pongWait);
+        if (pingInterval) clearInterval(pingInterval);
+        closed(code, reason);
+      });
+    });
+  }
+
+  async stop() {
+    for (const client of this.wss.clients) {
+      client.close(1001, 'Going away');
+    }
+    this.wss.removeAllListeners();
+    await new Promise<void>((resolve, reject) => {
+      this.wss.close((err) => (err ? reject(err) : resolve()));
+    });
+  }
+}
--- a/lib/graphql-ws/graphql-ws.exception.ts
+++ b/lib/graphql-ws/graphql-ws.exception.ts
@@ -1,0 +1,14 @@
+/**
+ * Base exception that closes a WebSocket connection when using `graphql-ws`.
+ */
+export class GraphqlWsException extends Error {
+  /**
+   * Instantiate a `GraphqlWsException`.
+   *
+   * @param reason Message explaining the error.
+   * @param code Code for a `CloseEvent`.
+   */
+  constructor(readonly reason: string, readonly code: number) {
+    super(reason);
+  }
+}
--- a/lib/graphql-ws/is-graphql-ws-subscription.util.ts
+++ b/lib/graphql-ws/is-graphql-ws-subscription.util.ts
@@ -1,0 +1,14 @@
+import {
+  GraphQLWsSubscriptionsConfig,
+  SubscriptionConfig,
+} from '../interfaces/gql-module-options.interface';
+
+export function isGraphQLWSSubscriptionConfig(
+  options?: SubscriptionConfig,
+): options is GraphQLWsSubscriptionsConfig {
+  return (
+    !!options &&
+    'protocol' in (options as any) &&
+    (options as any).protocol === 'graphql-ws'
+  );
+}
--- a/lib/graphql.factory.ts
+++ b/lib/graphql.factory.ts
@@ -26,6 +26,7 @@ import {
   ScalarsExplorerService,
 } from './services';
 import { extend, removeTempField } from './utils';
+import { isGraphQLWSSubscriptionConfig } from './graphql-ws/is-graphql-ws-subscription.util';
 
 @Injectable()
 export class GraphQLFactory {
@@ -50,6 +51,12 @@ export class GraphQLFactory {
       options.plugins || [],
       this.pluginsExplorerService.explore(),
     );
+
+    // remove subscriptions configuration if graphql-ws is used
+    if (isGraphQLWSSubscriptionConfig(options.subscriptions)) {
+      const { subscriptions, installSubscriptionHandlers, ...rest } = options;
+      options = { ...rest };
+    }
 
     const transformSchema = async (schema: GraphQLSchema) =>
       options.transformSchema ? await options.transformSchema(schema) : schema;
--- a/lib/graphql.module.ts
+++ b/lib/graphql.module.ts
@@ -33,6 +33,8 @@ import {
   mergeDefaults,
   normalizeRoutePath,
 } from './utils';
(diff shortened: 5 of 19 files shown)


Be-P:
I was reading the PR and it looks good to me, It would be great to switch to the `graphql-ws` library implementation.

asmeikal:
Hi @kamilmysliwiec, I've finally managed to do some thorough testing of this PR with the subscription services I'm working on, and they seem to work without problems (after upgrading the clients). I have tested this with an Electron application using React, and with two React Native applications. I was able to migrate the authentication (performed in the `onConnect` callback) with little effort, and everything else keeps working without changes. I've rebased the PR with this repo's master branch. I'm available to do some additional changes if necessary.

aurimukas:
Hey !
Any news for merging PR to default repo ??

jmcdo29:
Apologies for the delay. Currently we're focused on getting v8 out the door. Once it's published we will work on merging in PRs again. Kamil has been very busy with getting this going.

aurimukas:
> Apologies for the delay. Currently we're focused on getting v8 out the door. Once it's published we will work on merging in PRs again. Kamil has been very busy with getting this going.

Thanks for the answer.

kamilmysliwiec:
LGTM!

ddehghan:
Hi, Given that Apollo 3.0 has removed subscriptions-transport-ws completely we are suck. We would like to upgrade to graphql-ws but cant wait until nestjs 9.0 to be release. When do you expect that to go live? probably a year from now?

any change you can backport this to 8.0?

kamilmysliwiec:
I'll release v9 (exclusively for `@nestjs/graphql`) sooner to not keep everyone locked with v2.

asmeikal:
Hi @kamilmysliwiec! Thanks for merging this!

`graphql-ws` has received a major upgrade recently (see https://github.com/enisdenjo/graphql-ws/issues/117) that makes the version used in this PR not compatible with the latest version of the client. It should be enough to upgrade the dependency.

When we migrated from `subscriptions-transport-ws` to `graphql-ws` we ended up with something slightly different from this PR: we had to support both protocols, and ended up implementing something like this https://github.com/enisdenjo/graphql-ws#ws-backwards-compat. Maybe it's possible to do something similar here too? What is your roadmap for v9?

kamilmysliwiec:
@asmeikal this sounds great! Would you like to create a PR for this (both dep upgrade + backward compatibility)?

I'm hoping to get v9 ready asap to unblock people that want to use Apollo v3. I'll release it separately from other packages :)

asmeikal:
Hi @kamilmysliwiec, `graphql-ws` suggests to rely on websocket error codes to handle application level errors. The `1011` error code currently used on the `9.0.0` branch is a fatal error code ([see here](https://github.com/enisdenjo/graphql-ws/blob/3ffcd1ab57a48a7a90e651eb44894f0ac0db89d8/docs/interfaces/client.clientoptions.md#retryattempts) for a complete list), and the client would stop reconnecting when receiving it. The purpose of the custom exception was to allow custom error handling on the application level, e.g., to signal that the authorization token is expired and must be refreshed, without touching the socket directly.

Do you still need help with integrating the new `graphql-ws` version? If not, I can start testing it with our app!

kamilmysliwiec:
> The purpose of the custom exception was to allow custom error handling on the application level, e.g., to signal that the authorization token is expired and must be refreshed, without touching the socket directly

I removed the error handling layer because we no longer maintain the ws server internally but instead, we reuse the existing one (with `useServer`) and so there are no `error` and `onMessage` listeners (with try..catch blocks as in this PR). I figured that setting the error code manually (for example, `context.extra.socket.close(4000, 'Missing authorization');`) is sufficient but let me know if you think we should automatically wrap hooks like `onConnect` and hardcode error handling logic in there.

> If not, I can start testing it with our app!

If you can give it a shot, that would be awesome!

ddehghan:
OK we tested the 9.0.0-Next.2 package. it works. :-)

The main issue we have is the path is hardcoded to  '/graphql' . we need that to be configurable so that we can map it to something different so that on our load balancer we can set different timeout for each path. 

Haven't tested the exceptions yet. But will give that a try today

kamilmysliwiec:
Thanks for the feedback @ddehghan! Would you like to create a PR to address that?

ddehghan:
I did some more investigation:

Actually it seems that '/graphql' path has no effect only the domain portion is actually used.. The Client can request any url on the port and it gets connected. so there must be a bug somewhere else.

Because any url works we are able to set a different timeout of that load balancer so it works for us at the moment. web socket does support path. But I think the way they do it is they give you a call back on connection with the url and you are supposed to handle that. I don't think graphql-ws has implemented that so passing the path is a no-op.

asmeikal:
> > The purpose of the custom exception was to allow custom error handling on the application level, e.g., to signal that the authorization token is expired and must be refreshed, without touching the socket directly
> 
> I removed the error handling layer because we no longer maintain the ws server internally but instead, we reuse the existing one (with `useServer`) and so there are no `error` and `onMessage` listeners (with try..catch blocks as in this PR). I figured that setting the error code manually (for example, `context.extra.socket.close(4000, 'Missing authorization');`) is sufficient but let me know if you think we should automatically wrap hooks like `onConnect` and hardcode error handling logic in there.
> 
> > If not, I can start testing it with our app!
> 
> If you can give it a shot, that would be awesome!

Hi there, got around to testing the 9.0.0 branch!

About the ws server, the `useServer` provided by `graphql-ws` is intended to be an example implementation ([see here](https://github.com/enisdenjo/graphql-ws/blob/master/src/use/ws.ts#L29)) that can and should be updated based on the requirements of the integration. The initial implementation in this PR did change the example implementation to only allow ws connections on `/graphql`, as per @ddehghan first comment. I believe that the changes in this commit https://github.com/nestjs/graphql/commit/5641df041b091408f350ff05b7fa5552128c81c6 caused ws connections to any path to be passed to the ws server.

About the compatibility with `subscription-transport-ws`, I think some parts are missing: [here](https://github.com/nestjs/graphql/blob/9.0.0/lib/graphql-ws/graphql-ws-subscription.service.ts#L45) the subscription server should receive some onConnect/onDisconnect callbacks (to handle e.g. authentication), but they are missing. The big problem here is that they are typed really differently from the `graphql-ws` ones. Since this will be included in a major release, maybe the right approach would be to provide two keys in `GqlModuleOptions`: one for `graphql-ws` configuration and one for `subscription-transport-ws` configuration, and start only the servers for which the configuration was provided (neither, both, or just one of them). I can try and submit a PR for this change, if the interface I proposed is ok for you @kamilmysliwiec!

Last thing: about the error messages, I have strong opinions against using the socket directly. :) I believe that a custom WebSocket exception with a code for the WebSocket CloseEvent is the right choice, for a few reasons:
- the custom exception could be handled in different ways by `graphql-ws` and `subscription-transport-ws` in a transparent way for the developer that needs to support both protocols: `graphql-ws` would send the CloseEvent code to the client (which is what `graphql-ws` suggests), and `subscription-transport-ws` would send the exception message or some other string error code to the client
- if the developer has to use the socket directly, retrocompatibility with `subscription-transport-ws` is more difficult: the auth logic inside the `onConnect` would need to be heavily wrapped to support both protocols
- it's more similar to other exceptions that Nest gives to the developer, like the HttpException that contains a status code

kamilmysliwiec:
> I believe that the changes in this commit 5641df0 caused ws connections to any path to be passed to the ws server.

Great catch! Would you like to include a fix in your PR or do you want me to do so?

> About the compatibility with subscription-transport-ws, I think some parts are missing: here the subscription server should receive some onConnect/onDisconnect callbacks (to handle e.g. authentication), but they are missing

🕐  _todo_

> I can try and submit a PR for this change, if the interface I proposed is ok for you @kamilmysliwiec!

Sounds good!

> Last thing: about the error messages, I have strong opinions against using the socket directly. :)

Totally agree with you. I'm against using the socket directly as well, I'm just afraid of putting too much magic/hardcoding custom logic & auto-wrapping callbacks in this package. As implementing such an error layer doesn't really require changes in the `GraphQLModule` and could be implemented on the project's side (by library consumers), I figured _not having this_ available out-of-the-box shouldn't be a blocker.

Would you like to create 2 separate PRs to the v9 branch in which the latter one would contain errors layer-specific changes (so we can look at both separately)? 🙌

asmeikal:
Hi @kamilmysliwiec, I can try and put together two PRs for these changes! I hope to get at least one done by tomorrow!

kamilmysliwiec:
Amazing, thank you @asmeikal!

PR #1372, opened 2021-02-10
feat(input-type-factory): logs the `InputType` class name, if it is available

fregadev:
## PR Checklist

- [ x ] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

While building the GraphQL input types code first, a common mistake is to forget decorating the class with the `@InputType` decorator. When that happens, the user is currently presented with a message containing the input arg name, but not the input type that is missing.

Issue Number: N/A

## What is the new behavior?

When a user forgets to decorate an input type class, the message will show both the input arg name and the input type class name.

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/lib/schema-builder/errors/cannot-determine-input-type.error.ts
+++ b/lib/schema-builder/errors/cannot-determine-input-type.error.ts
@@ -1,7 +1,10 @@
+import { GqlTypeReference } from '../../interfaces';
+
 export class CannotDetermineInputTypeError extends Error {
-  constructor(hostType: string) {
+  constructor(hostType: string, typeRef?: GqlTypeReference) {
+    const inputObjectName: string | false = typeof typeRef === 'function' && typeRef.name;
     super(
-      `Cannot determine a GraphQL input type for the "${hostType}". Make sure your class is decorated with an appropriate decorator.`,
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
Can you please share the example how the final log message would look like in this case?

fregadev:
> Can you please share the example how the final log message would look like in this case?

Sure!

 It will look something like this:

`Cannot determine a GraphQL input type for the "find", on "PlannerFind". Make sure your class is decorated with an appropriate decorator.`

Given I have a mutation that looks like:
``` ts
  @Mutation(() => Planner)
  async find(
    @Args('find') find: PlannerFind,
  ): Promise<Planner> {
    return ...
  }
```

and an input:
``` ts
// @InputType() if we comment the decorator
export class PlannerFind {
  @Field(() => ID)
  _id: string;
}
```

kamilmysliwiec:
In this case, I'm not sure if "on" is a good preposition here. @jmcdo29 any ideas?

fregadev:
@kamilmysliwiec 
what about:

Cannot determine a GraphQL input type for the "find" with class name "PlannerFind".

Issue #1376, opened 2021-02-12
IntersectionType wrong types from generic InputType with array properties


notmedia:
## I'm submitting a...

<pre><code>
[ ] Regression 
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

When using IntersectionType with generic type which has property of array type, this type moves to single type.

## Expected behavior

The type should be with same properties as expected.

## Minimal reproduction of the problem with instructions

`where.args.ts`
```ts
import { Type } from '@nestjs/common';
import { Field, InputType } from '@nestjs/graphql';

export function WhereArgs<T>(classRef: Type<T>): any {
  @InputType({ isAbstract: true })
  abstract class WhereArgsType {
    @Field(() => [classRef], {
      description: 'All conditions must return true.',
      nullable: true,
    })
    AND?: T[];

    @Field(() => [classRef], {
      description: 'One or more conditions must return true.',
      nullable: true,
    })
    OR?: T[];

    @Field(() => [classRef], {
      description: 'All conditions must return false.',
      nullable: true,
    })
    NOT?: T[];
  }

  return WhereArgsType;
}
```

`get.args.ts`
```ts
import { ArgsType, Field, InputType, IntersectionType } from '@nestjs/graphql';

import { PaginationArgs } from '../../common/graphql';
import { WhereArgs } from '../../common/graphql/where.args';
import { StringFilter } from '../../common/graphql/where-filters/string.filter';

@InputType()
export class WhereEntityFields {
  @Field(() => StringFilter, {
    description: 'Entity name.',
    nullable: true,
  })
  name?: StringFilter;
}

@InputType()
export class WhereEntity extends IntersectionType(
  WhereEntityFields,
  WhereArgs(WhereEntityFields),
  InputType
) {}

@ArgsType()
export class GetEntityArgs extends PaginationArgs {
  @Field(() => WhereEntity, { nullable: true })
  where?: WhereEntity;
}
```

This generates `schema.gql`
```gql
input WhereEntity {
  """Entity name."""
  name: StringFilter

  """All conditions must return true."""
  AND: WhereEntityFields

  """One or more conditions must return true."""
  OR: WhereEntityFields

  """All conditions must return false."""
  NOT: WhereEntityFields
}
```

But I expect that with IntersectionType `schema.gql` should be

```gql
input WhereEntity {
  """Entity name."""
[...]

PR #1420, opened 2021-03-05
fix(@nestjs/graphql): support intersection-type for arrays

notmedia:
## PR Checklist

## PR Type

The problem was in `createWrappedExplicitTypeFn` function, when it executes in `reflectTypeFromMetadata` the options did not updated, because options modifying in returned callback from `createWrappedExplicitTypeFn`. But this options need to know before callback executes.

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

https://github.com/nestjs/graphql/issues/1376

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/lib/utils/reflection.utilts.ts
+++ b/lib/utils/reflection.utilts.ts
@@ -100,14 +100,14 @@ function createWrappedExplicitTypeFn(
   explicitTypeFn: ReturnTypeFunc,
   options: TypeOptions,
 ) {
-  return () => {
-    const explicitTypeRef = explicitTypeFn();
-    if (Array.isArray(explicitTypeRef)) {
-      const { depth, typeRef } = getTypeReferenceAndArrayDepth(explicitTypeRef);
-      options.isArray = true;
-      options.arrayDepth = depth;
-      return typeRef;
-    }
-    return explicitTypeRef;
-  };
+  let explicitTypeRef = explicitTypeFn();
+
+  if (Array.isArray(explicitTypeRef)) {
+    const { depth, typeRef } = getTypeReferenceAndArrayDepth(explicitTypeRef);
+    options.isArray = true;
+    options.arrayDepth = depth;
+    explicitTypeRef = typeRef;
+  }
+
+  return () => explicitTypeRef;
 }
--- a/tests/type-helpers/intersection-type.helper.spec.ts
+++ b/tests/type-helpers/intersection-type.helper.spec.ts
@@ -20,6 +20,9 @@ describe('IntersectionType', () => {
 
     firstName?: string;
 
+    @Field(() => [String])
+    hobbies?: string[];
+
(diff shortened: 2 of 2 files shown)


lib/utils/reflection.utilts.ts
```diff
@@ -100,14 +100,14 @@ function createWrappedExplicitTypeFn(
...
-    const explicitTypeRef = explicitTypeFn();
-    if (Array.isArray(explicitTypeRef)) {
-      const { depth, typeRef } = getTypeReferenceAndArrayDepth(explicitTypeRef);
-      options.isArray = true;
-      options.arrayDepth = depth;
-      return typeRef;
-    }
-    return explicitTypeRef;
-  };
+  let explicitTypeRef = explicitTypeFn();
```
kamilmysliwiec, line 103: As far as I remember, this line must be called inside `() =>` (lazily) to prevent circular-dependency-related issues.

notmedia: Can you provide me a case when it can cause the error?

I just tried this, and it works with this code. Or this is not the circular-dependency?

```js
  @ObjectType()
  class ClassA {
    @Field({ nullable: true })
    login: string;

    @Field()
    password: string;
  }

  @ObjectType()
  class ClassB {
    @Field()
    lastName: string;

    firstName?: string;

    @Field(() => [String])
    hobbies?: string[];

    @Field(() => [[ClassB]])
    matrix?: [ClassB[]];

    static [METADATA_FACTORY_NAME]() {
      return {
        firstName: { nullable: true, type: () => String },
      };
    }
  }
```

Btw I am try to rewrite this.

notmedia: Okay, then the solution would be to put the options variable in callback like get typeFn works. In this callback I can calculate options, like array depth etc. But as I see this will require a lot of rework. Can you advise better solution? Or I can start to do my variant.

notmedia: @kamilmysliwiec ?

notmedia: @kamilmysliwiec  I really need this fix ASAP, can you answer me?

notmedia: @kamilmysliwiec ?

notmedia: @kamilmysliwiec ?

notmedia: @kamilmysliwiec so, are you interested to fix this bug?

juansb827:
This is still happening 😞

notmedia:
@juansb827 Unfortunately the maintainers are not interested to fix this bug 😒

kamilmysliwiec:
I'll look into this as soon as I can. Let me lock this for now

PR #1473, opened 2021-04-02
feat(plugin): Update plugin with null union support

vinnymac:
## PR Checklist

- [ ] Docs have been added / updated (for bug fixes / features) **Where should I add the documentation for this?**

## PR Type
The GraphQL Plugin has been updated to broaden what code is considered nullable.

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Prior to these changes, the GraphQL plugin only supported decorating field properties when the type of a class property was optional.

Issue Number: N/A

## What is the new behavior?
Sometimes typescript users like to use union types that include null instead to indicate a nullable field. This change makes it possible to support nullable fields when the union includes a null type.

```
[x] Yes
[ ] No
```

Fields may now be nullable that were not before, this can be considered unexpected for the user.

## Other information

I noticed this functionality was missing while working with the NestJS GraphQL Plugin and my class properties such as `middleName: string | null;` were not being decorated as `nullable`. With this change, that can be supported.


kamilmysliwiec:
This sounds promising! I'll review it shortly. In the meantime, would you like to create a PR with similar functionality in the `@nestjs/swagger` repository? https://github.com/nestjs/swagger Steps should be pretty much the same

vinnymac:
Seems like similar functionality already exists in the swagger package. In fact, I believe a combination of the technique used in that project, and what I did here would be more appropriate. These changes improve this project more than the other, but I'll make PRs for both so we can discuss.

vinnymac:
@kamilmysliwiec let me know if there is anything else you'd like to see added here.

Thanks

kamilmysliwiec:
LGTM. Merged to v8.0.0 branch

PR #1662, opened 2021-07-27
feat(): add separate configuration for subscriptions

asmeikal:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

## What is the new behavior?

Allows separate configuration for the GraphQL subscription server: the developer can configure and enable both `graphql-ws` and `subscriptions-transport-ws`. Both protocols can be enabled at the same time.

```
[x] Yes
[ ] No
```

Applications using `subscriptions-transport-ws` (the default provided by Apollo Server) just need to update their configuration. They can optionally enable `graphql-ws` (which is safer and more robust) to start switching their clients to the new protocol.

## Other information


lib/interfaces/gql-module-options.interface.ts
```diff
@@ -44,8 +57,7 @@ export interface GqlModuleOptions
   executorFactory?: (
     schema: GraphQLSchema,
   ) => GraphQLExecutor | Promise<GraphQLExecutor>;
-  installSubscriptionHandlers?: boolean;
-  subscriptions?: GraphQLWsSubscriptionsConfig | boolean;
+  subscriptions?: SubscriptionConfig;
```
kamilmysliwiec, line 60: So the goal with preserving the `installSubscriptionHandlers` option was to remain backward compatible and avoid breaking changes. Basically, whenever someone sets `installSubscriptionHandlers` to `true`, we'd fall back to the default configuration (`subscription-transport-ws`). Would love your thoughts on that 🙌

adelinaenache:
I was really looking for this option! However, i'd like to set a few more options on my apollo client (in `useServer`), my biggest need being `execute`. Is there any reason to not allow such flexibility?

lib/interfaces/gql-module-options.interface.ts
asmeikal: Hi @kamilmysliwiec, I updated the PR, do you thing something like this could work? Setting `installSubscriptionHandlers` enables `subscription-transport-ws` without any specific configuration. I imagine this is useful for subscriptions that don't need any authorization nor additional logic.

kamilmysliwiec: Sounds great @asmeikal! Thank you

kamilmysliwiec:
PR looks great, thank you @asmeikal!

PR #1751, opened 2021-09-18
WIP: Feature/update ts transform plugin

thekip:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number:https://github.com/nestjs/graphql/issues/1346

## What is the new behavior?
This PR brings changes in how current ts transform plugin is working. 

To Do:
- [x] Option to work in isolatedModules (no typeChecker available)
- [x] Introspection for `@deprecated` jsdoc tag
- [ ] Support for GraphQL scalar types: Int, Float, ID (please refer to the issue https://github.com/nestjs/graphql/issues/1346 i rised a discussion there)
- [ ] Support more complicated cases in no-typechecker mode.
- [ ] More tests covering edge cases in no-typechecker mode. 

## What was changed internally: 
1. JSDoc Comments now parsed using Typescript AST (built-it typescript support) instead of regular expressions 
2. Removed code related to imports and merging existing factories (please see linked issue)
3.  Test rewritten, Jest Snapshots used

## Does this PR introduce a breaking change?
```
[x] Yes
[ ] No
```
There is a defenetely changes which can impact users. 
TODO: eliminate them or at least find and clearly state what can go wrong. 

## Other information

Issue #1346, opened 2021-01-26
Are there any plans of extending GraphQL CLI Plugin functionality?


thekip:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

Right now plugin automatically can extract metadata only from entities, replacing `@Field` decorator functionality. 

There are some room of improvment, for example: 

- Automatic extraction jsdoc `@deprecated` tag and putting it as deprecation reason for schema
- Creating Opaque Types for overcoming inconsistency between JS/TS `number` and GraphQL's `Float`, `Int` 
  
```typescript
type Float = Opaque<number>;
type Int = Opaque<number>;

@ObjectType()
export class Launch {
  public id: Float;
}

 ``` 
 Type automatically extracted for schema as `Float` and still available in Typescript as `number`
- Expanding functionality of plugin  to Resolvers
  
```typescript
// instead
@Query(returns => [Launch], {
  description: 'Some description',
  deprecationReason: 'Becusase it\'s to much boilerplate!',
})
public async launches() {
  return this.launchApi.getAllLaunches();
}

// to

/**
 * Some description
 * @deprecated Becusase it's to much boilerplate!
 */
@Query() public async launches(): Promise<Launch[]> {
  return this.launchApi.getAllLaunches();
}
 ```
- to arguments

```typescript
@Query()
public async launch(@Arg() id: Int) {
  // (or even omit  @Arg() decorator at all, if you don't need to override defaults)
  return this.launchApi.getLaunchById(id);
}
  ```

I have an experience of AST transformations and Typescript, and we've just started a new project using NestJs + GraphQL, so have an ability to put some effort into this. But the question is what maintainers think about that? We can discuss a plan and how API should looks like.

kamilmysliwiec:
Some of these features are on my radar and I'll definitely look into implementing them in the future, but currently, I don't have enough time to focus on this, unfortunately. PRs are more than welcome though so if you are interested in contributing, feel free to create one 🙌

th3rius:
Let me know if you're working on this, I'm also willing to make a PR

kamilmysliwiec:
@th3rius there's no PR in progress so you're more than welcome to contribute!

thekip:
[...]

PR #1757, opened 2021-09-22
New features for GraphQL typescript plugin

thekip:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently GraphQL Plugin does not understand `@deprecated` tag in JsDoc comments for fields. 

Issue Number: https://github.com/nestjs/graphql/issues/1346

## What is the new behavior?
This PR adds support for `@deprecated` tag. It parses it in following fashion: 

1. If just tag is presented, with no message then 'deprecated' default message would be used
2. If user specify a deprecation message after a tag:
```js
  /**
  * @deprecated consult docs for better alternative!
  */
  breed: string;
```
Message would be used. 

I'm using built-in typescript jsdoc support (may be need to check from what version typescript started support it) to parse jsdoc and get information of tags. 

```
[ ] Yes
[x] No
```
There are no general breaking changes, but schema might be fulfilled with unintended deprecation messages, and this could be quite surprising for some.

## Other information

This PR is rethinking of https://github.com/nestjs/graphql/pull/1751 Which is affecting bigger scope and might introduce breaking changes. I decided to roll out features one by one in small PR's instead.

Stay updated, i have many new useful features in my plan :)


thekip:
## About refactoring

Generally using all `ts.createXYZ` method of the compiler is deprecated. We should use corresponding methods from `NodeFactory` passed as context to transformation eq 'ctx.factory.createXYZ'

Some of the methods has 1 to 1 replacement, some of them have slightly changed name and few no more exists (createLiteral(string | bolean | number), createMutableClone, and etc)

thekip:
@kamilmysliwiec any feedback on this?

kamilmysliwiec:
I'll take a look as soon as I can

thekip:
3 more features added: 

1. Auto discover name in `registerEnumType()`

```ts
// previously
registerEnumType(MyAwesomeEnum, {name: 'MyAwesomeEnum' })

// with this feature (name is inhereted from Enum name)
registerEnumType(MyAwesomeEnum)
```

2. Auto discover name for `createUnionType()`
```ts
// previously
cont MyAwesomeUnion = createUnionType({name: 'MyAwesomeUnion', types: [Foo, Bar]})

// with this feature (name is inhereted from variable name)
cont MyAwesomeUnion = createUnionType({types: [Foo, Bar]})

```

3. Auto discover enums (disabled by default). With this feature enabled all enums in processed files would be automatically exposed to graphql schema (registered with registerEnumType).  All comments introspection features such as description and deprecation works for enums as well. 
```ts
/** This enum represents possible statuses */
enum Status {
  /** This is status by default */
  ENABLED,
  DISABLED,
  /** @deprecated please use other options */
  OTHER,
}

// This will produce under the hood: 

registerEnumType(Status, {
  name: 'Status',
  description: 'This enum represents possible statuses',
  valuesMap: {
    ENABLED: {
      description: 'This is status by default'
    },
    OTHER: {
      deprecationReason: 'please use other options',
    }
  }
})
```

If you want to exclude particalure enum from exposing you need to add `@private` or `@HideEnum` jsDoc tag (the tags TBD).

```ts
/**
* @private
*/
enum Status {
    ENABLED,
    DISABLED
}
// OR

/**
* @HideEnum
*/
enum Status2 {
    ENABLED,
    DISABLED
}
```

kamilmysliwiec:
This is looking great so far @thekip, thank you! I left a few minor comments so far but I'll definitely pull all changes and walk through them more meticulously very soon.

thekip:
Do nestjs still support typescript version prior 4.2? Where i can find a minimal system requirements?

thekip:
Well, for such things as a typescript transformation plugin we should define a "minimal version" because we a heavily relay on typescript compiler API which changes over time. So at least for the plugin we should define it and state it clearly in the docs (something like if you want to use typescript plugin you should have typescript > 4.2) 

It also would be nice to have to setup test matrix with different typescript versions on different OS (i saw branches in source code which relay on different path windows/unix). 

## Side topic Error handling in transformation 

Right now there are no such thing as error handling. If a transformation could not figure out a type for property it's just silently leave `() => Object` instead of type. This doesn't have any sense for runtime consumer and it shows error while trying to generate a scheme. Usually errors from scheme generator is quite inaccurate and may point user in wrong direction. 

During compiling we have more contextual information (for example inline type literal is used) and we can generate more accurate errors which would point to exact place in source file. 

Example:

> GraphqlPluginError: Property `test` in `CatDto` class has unsupported type (TypeLiteralExpression). __ additional message helping to solve the issue __

The main problem here is how to throw this exceptions. We could not just throw it from the transformation because it will exit a typescript compilation. (not a desirable behaviour in watch mode). Any way need to investigate on topic "How to properly report errors from typesctipt transform" 

For me obvious option is generate `throw new Error(...)` and insert it into sourcecode and throwing errors from the runtime. (errors should be inserted in top level of a module, to be executed as early as possible) 

What do you think? Does it make any sense?

kamilmysliwiec:
> Well, for such things as a typescript transformation plugin we should define a "minimal version" because we a heavily relay on typescript compiler API which changes over time. So at least for the plugin we should define it and state it clearly in the docs (something like if you want to use typescript plugin you should have typescript > 4.2)

> For me obvious option is generate throw new Error(...) and insert it into sourcecode and throwing errors from the runtime. (errors should be inserted in top level of a module, to be executed as early as possible)

Both suggestions sound good to me

thekip:
## New feature

Plugin now can introspect type from getters: 

```typescript
@ObjectType()
class ObjectTypeModel {
  get prop(): string {
    return 'something';
  }
}
```

kamilmysliwiec:
Would you be able to rebase your changes @thekip? If you're busy, I can look into this instead

thekip:
I will

thekip:
@kamilmysliwiec i synchonized the branch with current master. Unfortunately i was not able to rebase and keep branch clean, due to file structure changes and re-formatting files. I hope i didn't miss anything. 

Please take a look, and let's discuss future of the branch and typescript plugin overall.

thekip:
I also have quite interesting and 'brave' ideas how to improve DX with code-first approach avoiding sharp edges with typescript compiler and etc. I would like to share it with you because scope of changes might be quite big.

thekip:
@kamilmysliwiec if some features are controversial, we can drop them and merge only those which are more predictable and can bring more value.

thekip:
I linked forked plugin to our app, and found an issue introduced with this commit  https://github.com/nestjs/graphql/pull/1757/commits/3fe7cdbdf1118a105f44df44fde7329f55817d38

Looks like something was changed in the mechanism of merging factories provided from the cli plugin and runtime decorators. Latest commit fixes the issue. Now it works on our moderate-sized codebase.

kamilmysliwiec:
I'd love to merge this PR but I feel like we're trying to bring too many features at once. If we could cherry-pick the following changes for now:

1. Support for the `@deprecated` tag
2. Migration to use `ctx.factory` instead of `ts.[..]`
3. Support extracting description on a type level

that would be fantastic! Afterward, we could think about changes around enums/unions (preferably in separate PRs - for example - auto-generating implicit enums for inline string unions sounds very useful as well and I'd love to get this one released under a flag for now). I understand that this (reverting other changes) might be somewhat time-consuming so let me know if you're willing to work on this one and I can try to cherry-pick the rest by myself! @thekip

thekip:
Ok, i will take care of that

kamilmysliwiec:
I went ahead and removed features that I've mentioned here https://github.com/nestjs/graphql/pull/1757#issuecomment-1228416518 Feel free to create separate PRs with those individual features 🙌 

Fantastic PR, thanks for your contribution!

EdouardBougon:
Since this new version, it seems that the graphQL plugin doesn't introspect anymore the fields on `Args` class.

I just updated my code with the version, and all field on args disappear from the graphQL schema.

Anyone with the same issue ?

kamilmysliwiec:
@EdouardBougon could you create a new issue with a minimum reproduction repository?

kamilmysliwiec:
@EdouardBougon this line likely causes this issue https://github.com/nestjs/graphql/blob/master/packages/graphql/lib/plugin/visitors/model-class.visitor.ts#L33

Can you modify this array locally (adding ArgsType) and check if it fixes your issue?

EdouardBougon:
Doing it right now

EdouardBougon:
yes I confirm, it fix the issue

EdouardBougon:
@kamilmysliwiec Do you still need a repo to test the issue ?

Issue #1238, opened 2020-11-10
GqlParamsFactory isn't compatible with Federation


ThisIsMissEm:
## I'm submitting a...

<pre><code>
[ ] Regression 
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

In Apollo Federation, resolvers for references are called with only three arguments, rather than the standard four, which in the current implementation of GqlParamsFactory results in the `ROOT` being the `ARGS` and so fourth, as all the arguments are shifted one to the left.

This results in the arg decorator trying to fetch the argument from the GraphQL `context` object instead of the `args` object (array?) and the context decorators trying to fetch from the `info`.

## Expected behavior

GqlParamsFactory should be Apollo Federation aware, and correctly handle `args` being of length 3 for referenceResolvers and 4 for normal resolvers. e.g., there'd be a conditional here:

```
case GqlParamtype.CONTEXT:
        return data && args[2] ? args[2][data as string] : args[2];
```

To be something like (not exactly that, but to show the issue/idea):

```
case GqlParamtype.CONTEXT:
        return data && args[args.length === 3 ? 1 : 2] ? args[args.length === 3 ? 1 : 2][data as string] : args[args.length === 3 ? 1 : 2];
```

## Minimal reproduction of the problem with instructions

No easy way to prepare a minimal reproduction, unfortunately.

## What is the motivation / use case for changing the behavior?

Have everything work as expected. Currently we have this work-around in our `createParamDecorator` for our "token" param/arg for if that decorator is used on a resolveReference field:

```
    /*
     * A GqlExecutionContext typically expects 4 args in ExecutionContext.
     * However, in a @ResolveReference context, there are 3 args: `reference`, `context`, and `info`.
     * (See https://www.apollographql.com/docs/federation/api/apollo-federation/#parameters-1)
     *
     * This causes a problem, because GqlExecutionContext expects the first argument to be
     * the value of the `rootValue` function. (See https://www.apollographql.com/docs/apollo-server/data/resolvers/#resolver-arguments)
     *
     * So, in a hacky maneuver, we fix the args indices for the @ResolveReference scenario by shifting them one to the right.
     */
    if (context.getArgs().length < 4) {
      context.getArgs().unshift(undefined);
    }
```

## Environment

<pre><code>
Nest version: 7.0.0 (issue is visible on master branch)

 
[...]

PR #1996, opened 2022-02-09
fix(graphql): Fix federation reference resolver context args

mohatt:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #1238

Currently, when trying to use a request-scoped provider (that injects graphql context) inside a federation reference resolver, its initiated with a wong context instance due to the fact that reference resolvers are invoked with only 3 arguments. That's because `__resolveReference` isn't a `GraphQLFieldResolver`, it's a `GraphQLReferenceResolver` ([as defined in @apollo/subgraph](https://github.com/apollographql/federation/blob/947948f9fd37a64ec3b08d181ffa0c9fd7bda223/subgraph-js/src/schemaExtensions.ts#L3)).

```typescript
import { Inject, Injectable, Scope } from '@nestjs/common'
import { CONTEXT } from '@nestjs/graphql'

@Injectable({ scope: Scope.REQUEST })
export class ProductLoader {
  constructor(@Inject(CONTEXT) private context: any) {
   // context here is not a valid graphql context (its basically the graphql `info` argument)
   console.log(context)
  }
  
  load(id) {
    console.log(id)
  }
}

@Resolver(() => Product)
export class ProductResolver {
  constructor(private loader: ProductLoader) {}

  @ResolveReference()
  async resolveReference(reference) {
    return this.loader.load(reference.id)
  }
}
```

## What is the new behavior?
This pr fixes the issue by adding a root argument with `undefined` value to `args` array when its length is less than four, I am not sure if thats the correct way to fix this but it seems to be working.

I would love to add tests for this pr but would need some guidence on where to add them.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/graphql/lib/factories/params.factory.ts
+++ b/packages/graphql/lib/factories/params.factory.ts
@@ -7,6 +7,10 @@ export class GqlParamsFactory implements ParamsFactory {
     if (!args) {
       return null;
     }
+    // Reference resolver args don't have root argument
+    if (args.length === 3) {
+      args = [undefined, ...args];
+    }
     switch (type as GqlParamtype) {
       case GqlParamtype.ROOT:
         return args[0];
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
LGTM

PR #1432, opened 2021-03-10
fix(): support multiple GraphQLModule in an app

wtho:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [X] Tests for the changes have been added
- [ ] Docs have been added / updated *Docs were already formulated as if this bug didn't exist*

## PR Type
What kind of change does this PR introduce?

```
[X] Bugfix
```

## What is the current behavior?
Currently, when wanting to use multiple GraphQLModules (different schemas for separate endpoints) in combination with the code-first approach, there will be an error such as `Error: Schema must contain uniquely named types but contains multiple types named "User".`.

Issue Number: #721

## What is the new behavior?
Now it works.
During `GraphQLModule` initialization, basically all storages are cleared and the codebase is re-scanned from scratch including re-evaluation of annotations. Only the `TypeDefinitionsStorage` is an exception, which keeps references to outdated objects using the maps `InputTypeDefinitionsLinks` and `outputTypeDefinitionsLinks`.

With the new behavior, these storages are cleared as well in the `GraphQLSchemaFactory` during `GraphQLModule` initialization.

## Does this PR introduce a breaking change?
```
[ ] Yes
[X] No
```

## Other information
I opted to keep code changes minimal while fixing the bug. This means I mixed static method calls with intance method calls in the `GraphQLSchemaFactory` and extended an existing test to mimic the problem. If you want me to add an additional test or structure the code differently, just let me know.
Same goes for a Changelog or docs, so far I haven't touched any.

Closes #721

Issue #721, opened 2020-03-30
GraphQl Multiple Endpoints not working 


arslanmughal5566:
## Current behavior
When adding multiple endpoints for graphql it gives following weird error 
```
(node:16316) UnhandledPromiseRejectionWarning: Error: Schema must contain uniquely named types but contains multiple types named "User".
    at typeMapReducer (C:\Users\arsla\Documents\nest-graphql-practice\node_modules\graphql\type\schema.js:262:13)
    at Array.reduce (<anonymous>)
    at new GraphQLSchema (C:\Users\arsla\Documents\nest-graphql-practice\node_modules\graphql\type\schema.js:145:28)
    at GraphQLSchemaFactory.<anonymous> (C:\Users\arsla\Documents\nest-graphql-practice\node_modules\@nestjs\graphql\dist\schema-builder\graphql-schema.factory.js:28:28)
    at Generator.next (<anonymous>)
    at C:\Users\arsla\Documents\nest-graphql-practice\node_modules\tslib\tslib.js:113:75
    at new Promise (<anonymous>)
    at Object.__awaiter (C:\Users\arsla\Documents\nest-graphql-practice\node_modules\tslib\tslib.js:109:16)
    at GraphQLSchemaFactory.create (C:\Users\arsla\Documents\nest-graphql-practice\node_modules\@nestjs\graphql\dist\schema-builder\graphql-schema.factory.js:23:24)
    at GraphQLSchemaBuilder.<anonymous> (C:\Users\arsla\Documents\nest-graphql-practice\node_modules\@nestjs\graphql\dist\graphql-schema.builder.js:55:56)
(node:16316) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:16316) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
## Input Code
My appModule look the following 

```ts
import { Module } from '@nestjs/common';
import { GraphQLModule } from '@nestjs/graphql';
import { MessagesModule } from './messages/messages.module';
import { UserModule } from './user/user.module';

@Module({
  imports: [
    MessagesModule,
    UserModule,
    
    GraphQLModule.forRoot({
      autoSchemaFile: 'userSchema.gql',
      playground: true,
      include: [UserModule],
[...]

PR #1999, opened 2022-02-10
Allow for multiple graphql endpoints in the same application with schema segregation

Aetherall:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## Why this feature ?

In some approaches of application development, we want to split our code into bounded contexts.
GraphQL's federation allows us to defer the aggregation of data needed for cross-contexts read use-cases.  
However, deployment of applications should be decoupled from their internal context dependency scheme,
as it allows flexibility throughout the discovery of the infrastructure requirements of each context. 

Currently, having 2 `GraphQLModule` in the same application will result in the application not wanting to start, yielding the following error :
```
Schema must contain uniquely named types but contains multiple types named "User".
```

This is due to the fact that we add GraphQL types/interface/etc. to the schema globally.
Cf: [`orphaned-types.factory.ts`](https://github.com/nestjs/graphql/blob/bd71506d362dd2f73ba281c6cf7e4f111a622737/packages/graphql/lib/schema-builder/factories/orphaned-types.factory.ts)

This have the effect to generate schemas containing definitions for both contexts,
thus creating a conflict beween original `ObjectTypes` and their extended counterparts named the same.

## What is the new behavior?

The new behavior would allow for a nest user to have multiple GraphQL endpoints in a single nest application.
Those endpoints would expose a schema only containing definitions referenced by the resolvers below the included modules.
An example application, along with corresponding tests, can be found in the apollo package of the repository.

- [ ] Yes
- [ ] No
- [x] Not known yet

## Other information

I am asking for insights on how I should approach the problem.
The newly written tests are passing when entirely disabling the orphaned type retrieval ^^'
It could be useful to me to understand the origin for global orphaned type registration, and if there is a way to access them from the resolvers used to create the schema.

Thanks a lot !

### DIFF
--- a/packages/apollo/tests/code-first-multigraph/app.module.ts
+++ b/packages/apollo/tests/code-first-multigraph/app.module.ts
@@ -1,0 +1,8 @@
+import { Module } from '@nestjs/common';
+import { PostModule } from './post/post.module';
+import { UserModule } from './user/user.module';
+
+@Module({
+  imports: [PostModule, UserModule],
+})
+export class ApplicationModule {}
--- a/packages/apollo/tests/code-first-multigraph/gateway.module.ts
+++ b/packages/apollo/tests/code-first-multigraph/gateway.module.ts
@@ -1,0 +1,18 @@
+import { Module } from '@nestjs/common';
+import { GqlModuleOptions, GraphQLModule } from '@nestjs/graphql';
+import { ApolloGatewayDriver, ApolloGatewayDriverConfig } from '../../lib';
+
+@Module({
+  imports: [
+    GraphQLModule.forRoot<ApolloGatewayDriverConfig & GqlModuleOptions>({
+      driver: ApolloGatewayDriver,
+      gateway: {
+        serviceList: [
+          { name: 'user', url: 'http://localhost:3000/user/graphql' },
+          { name: 'post', url: 'http://localhost:3000/post/graphql' },
+        ],
+      },
+    }),
+  ],
+})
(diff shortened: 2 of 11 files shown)


Aetherall:
I am wondering if the approach of exploring resolver metadata to reference the types will cover all use cases

kdawgwilk:
This is exactly the use case I was trying to attempt in #2164 however with schema first approach instead of code first

kdawgwilk:
It was mentioned in this PR https://github.com/nestjs/graphql/pull/1432#issuecomment-800981372 
>introduce a new feature that lets you associate types (e.g., object types) with NestJS modules, for example, using the following construction `@ObjectType({ registerIn: CatsModule })`. In this case, types wouldn't be global anymore and we could store them per app instead of a per process

Maybe that would be a direction to start in that would be supported by the maintainers?

Aetherall:
Thanks for the heads up @kdawgwilk, I was using my dirty fix since then and forgot about the issue.

I will try to implement something as soon as I get back from holidays. 

Don't hesitate to use my PR as an iteration base if you want to try something until then !

wodCZ:
I believe this is related to #721. Linking this for reference.

dontwan:
Any progress on this PR? 
I will also like to use multiple endpoints

kamilmysliwiec:
https://github.com/nestjs/graphql/pull/3811

PR #2072, opened 2022-03-11
feat(): add export for cli plugin for ts-patch/ttypescript

CarsonF:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

[ttypescript](https://github.com/cevek/ttypescript) & [ts-patch](https://github.com/nonara/ts-patch) allow for TS transformation similar to what nest's cli plugins offer.

Nest's configuration falls short in that it only applies when running through the nest cli command - an obvious limitation which the library shouldn't try to solve.

Similar to what is documented for configuring these plugins for jest, a custom file has to be created that conforms to the signature the caller is expecting.

For ttypescript/ts-patch, that looks like this
```ts
import { before } from '@nestjs/graphql/dist/plugin';
import { Program, TransformerFactory } from 'typescript';

export default (program: Program, config?: any): TransformerFactory<any> =>
  before(config, program);
```
And registered in `tsconfig.json`
```json
{
  "compilerOptions": {
    "plugins": [
      {
        "transform": "./tools/nestjs-graphql.transformer.ts",
        "typeFileNameSuffix": [".input.ts", ".args.ts", ".entity.ts", ".model.ts"],
        "introspectComments": true
      }
    ]
  }
}
```

As you can see the signature is very close, it just has the arguments order swapped. 

## What is the new behavior?

I added a default export with the arguments order swapped so that users don't need to add their own transformer file.
```diff
-        "transform": "./tools/nestjs-graphql.transformer.ts",
+        "transform": "@nestjs/graphql/plugin",
```

The benefit of using this approach is that the plugin is applied to nest cli, jest, ts-node, etc. all from the single configuration.

- [ ] Yes
- [x] No


kamilmysliwiec:
Sounds good to me!

PR #2131, opened 2022-04-11
fix: resolve type context args in request scoped providers

Davide-Gheri:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Importing a request scoped provider in a Resolver that defines a `__resolveType` decorated with `@ResolveField()` wrongly creates multiple provider instances in a single request, this is because a resolveType function has not the same arguments as a normal field resolver function (it is missing the `arg` argument), thus the request id used to keep track of the request is wrongly stored in the `info` object instead of the `context` object.
https://github.com/nestjs/graphql/blob/master/packages/graphql/lib/services/resolvers-explorer.service.ts#L170

Issue Number: N/A

## What is the new behavior?

This PR adds an `undefined` argument where the `args` object should be for `__resolveType` field resolvers

- [ ] Yes
- [x] No

### DIFF
--- a/packages/graphql/lib/factories/params.factory.ts
+++ b/packages/graphql/lib/factories/params.factory.ts
@@ -1,16 +1,16 @@
 import { ParamData } from '@nestjs/common';
 import { ParamsFactory } from '@nestjs/core/helpers/external-context-creator';
 import { GqlParamtype } from '../enums/gql-paramtype.enum';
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
LGTM

PR #2145, opened 2022-04-20
GraphQL Yoga driver

charlypoly:
This PR add supports for GraphQL Yoga which brings:
- Express support
- Fastify support
- embedded GraphiQL
- File uploads support
- CORS default
- Error masking
- subscriptions support
- lightweight bundle 📦 🪶 

https://www.graphql-yoga.com/

### DIFF
--- a/packages/graphql/lib/services/gql-subscription.service.ts
+++ b/packages/graphql/lib/services/gql-subscription.service.ts
@@ -1,4 +1,8 @@
-import { execute, GraphQLSchema, subscribe } from 'graphql';
+import {
+  execute as graphqlExecute,
+  GraphQLSchema,
+  subscribe as graphqlSubscribe,
+} from 'graphql';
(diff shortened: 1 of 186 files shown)


kamilmysliwiec:
@charlypoly thank you for this contribution, this is amazing! 

However, the reason we extracted drivers out of the "core" package is that we wanted to expose a mechanism that would let other community members integrate custom drivers that we _can't_ maintain & keep under the official organization (mainly due to the limited team bandwidth & time). Bringing another driver in would decrease our velocity even further (and given the amount of packages & drivers we already maintain, it's getting more and more difficult).

So in this case, I'd rather suggest creating a new repository & new package and publishing it as a separate community-driven OSS package. Once it's published, definitely let me know and I'll make sure we mention it (add a dedicated section in README/maybe docs too?) and link it here :)

PR #2201, opened 2022-05-26
fix(graphql): required exports for testing

charlypoly:
Hi @kamilmysliwiec,

In the motivation to implement the GraphQL Yoga Driver in a separate repository (previous discussion: #2145), we need the following exports to be available outside of the package for testing purposes: `GRAPHQL_SDL_FILE_HEADER` (schema generation test), `FileSystemHelper` (mocking for schema DSL file loading).

Example of an Apollo test that won't work if moved outside of this monorepo:

https://github.com/nestjs/graphql/blob/42704556db2e492b218a98091d97bfb3bbca607f/packages/apollo/tests/e2e/graphql-sort-auto-schema.spec.ts#L1-L8

### DIFF
--- a/packages/graphql/lib/index.ts
+++ b/packages/graphql/lib/index.ts
@@ -7,6 +7,7 @@ export * from './graphql-schema.host';
 export * from './graphql-types.loader';
 export * from './graphql.factory';
 export * from './graphql.module';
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
LGTM

PR #2202, opened 2022-05-27
refactor(graphql): add abstract context id factory

thiagomini:
change resolvers-explorer service to use an abstract contextId factory

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

The [resolvers-explorer.service](https://github.com/nestjs/graphql/blob/890762d532c3824037fcfc2c1ef81c653bbd14e7/packages/graphql/lib/services/resolvers-explorer.service.ts#L175) currently isn't using the `ContextIdFactory#getByRequest` method to retrieve the contextId from the request. Instead, it is manually checking the `gqlContext` object and calling the `createContextId()` function. This is a problem if we plan to monkeypatch the `ContextIdFactory` to create the separate Dependency Trees.

Issue Number: N/A

## What is the new behavior?

Now, `ContextIdFactory` is an abstract class and we have a GraphQL implementation for it. Although this class is defined in the graphql package, ideally it should reside in the `common` or `core` from the main repo. I only added it here to showcase what we could do.

Now, if we want to provide a specific implementation of ContextIdFactory for our needs, we can override it by ourselves.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/graphql/lib/factories/context-id-factory.ts
+++ b/packages/graphql/lib/factories/context-id-factory.ts
@@ -1,0 +1,28 @@
+import { ContextId, createContextId } from '@nestjs/core';
+import { REQUEST_CONTEXT_ID } from '@nestjs/core/router/request/request-constants';
+
+export abstract class ContextIdFactory {
+  abstract create(): ContextId;
+
+  abstract getByRequest<T extends Record<any, any> = any>(
+    request: T,
+  ): ContextId;
+}
+
+export class GraphqlContextIdFactory implements ContextIdFactory {
+  create(): ContextId {
+    return createContextId();
+  }
+  getByRequest<T extends Record<any, any> = any>(gqlContext: T): ContextId {
+    if (!gqlContext) {
+      return this.create();
+    }
+    if (gqlContext[REQUEST_CONTEXT_ID as any]) {
+      return gqlContext[REQUEST_CONTEXT_ID as any];
+    }
+    if (gqlContext.req && gqlContext.req[REQUEST_CONTEXT_ID]) {
+      return gqlContext.req[REQUEST_CONTEXT_ID];
+    }
+    return createContextId();
+  }
+}
--- a/packages/graphql/lib/services/resolvers-explorer.service.ts
+++ b/packages/graphql/lib/services/resolvers-explorer.service.ts
@@ -1,7 +1,6 @@
 import { Inject, Injectable, Logger } from '@nestjs/common';
 import { isUndefined } from '@nestjs/common/utils/shared.utils';
 import {
-  createContextId,
   MetadataScanner,
   ModuleRef,
   ModulesContainer,
@@ -23,6 +22,10 @@ import { SubscriptionOptions } from '../decorators/subscription.decorator';
 import { AbstractGraphQLDriver } from '../drivers/abstract-graphql.driver';
 import { GqlParamtype } from '../enums/gql-paramtype.enum';
 import { Resolver } from '../enums/resolver.enum';
+import {
+  ContextIdFactory,
+  GraphqlContextIdFactory,
+} from '../factories/context-id-factory';
 import { GqlParamsFactory } from '../factories/params.factory';
 import {
   FIELD_RESOLVER_MIDDLEWARE_METADATA,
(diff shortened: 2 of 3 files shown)


packages/graphql/lib/services/resolvers-explorer.service.ts
```diff
@@ -52,6 +54,8 @@ export class ResolversExplorerService extends BaseExplorerService {
     @Inject(GRAPHQL_MODULE_OPTIONS)
     private readonly gqlOptions: GqlModuleOptions,
     private readonly moduleRef: ModuleRef,
+    @Inject(GraphqlContextIdFactory)
```
thiagomini, line 57: In a final version, this should be removed in favor of the `ContextIdFactory` token, so we can provide an alternative Factory by ourselves using something like:
```ts
const myCustomContextIdFactoryProvider = {
  provide : GraphlqlContextIdFactory,
  useClass: MyCustomClass
}
```

packages/graphql/lib/services/resolvers-explorer.service.ts
```diff
@@ -172,24 +176,15 @@ export class ResolversExplorerService extends BaseExplorerService {
...
-          Object.defineProperty(gqlContext, REQUEST_CONTEXT_ID, {
-            value: contextId,
-            enumerable: false,
-            configurable: false,
-            writable: false,
-          });
-        }
+
+        const contextId: ContextId =
+          this.contextIdFactory.getByRequest(gqlContext);
```
thiagomini, line 181: I realized that this part of the code looked like [these](https://github.dev/nestjs/nest/blob/1ae72e1fdb338a6e00ac0ca19615602a91edc5bd/packages/core/middleware/middleware-module.ts#L212) lines from `middleware-module` in the core package. In the end, the intention of this part of the code was apparently to exactly the contextId from the request (in this case, the graphql context)

packages/graphql/lib/factories/context-id-factory.ts
```diff
@@ -0,0 +1,28 @@
+import { ContextId, createContextId } from '@nestjs/core';
+import { REQUEST_CONTEXT_ID } from '@nestjs/core/router/request/request-constants';
+
+export abstract class ContextIdFactory {
```
thiagomini, line 4: The reason why I use an abstract class here instead of an interface is to be able to use it as an Injection Token.

packages/graphql/lib/services/resolvers-explorer.service.ts
kamilmysliwiec: @thiagomini instead of refactoring the existing code to use the injected factory, could we simply use the `ContextIdFactory` static class from the `@nestjs/core`? As I've mentioned on the call, although it's feasible for the `@nestjs/graphql` package, we can't really rely on the container in this case as in microservices & router-explorer (REST), we construct routes outside of the container (so at the end of the day, even if we introduce "custom context id factory" feature, it will have to be applied **outside** of the DI container anyway).

Otherwise, fantastic job Thiago!

thiagomini: Hmmm yea, we could use the static ContextIdFactory here, but the problem is the implementation of the `getByRequest()` from `ContextIdFactory` would not work here, since it was created with the RESTful context in mind. If we look at the [implementation](https://github.dev/nestjs/nest/blob/7ac42b08ef31350d51b87d618be333fc69c1414f/packages/core/helpers/context-id-factory.ts#L28), it is trying to retrieve the `CONTEXT_ID` directly from the `request` object or `request.raw`, while in the graphql context we are trying to retrieve from the `request` or `request.req`.

With the above in mind, we could try to import the ContextIdFactory and override its implementation for `getByRequest` method, I suppose? But I feel this would be a work-around, instead of a long-term solution 🤔 I have another suggestion, since we can't have this injected via the DI container in some cases: 

1. We maintain the factory as a private attribute for this class, also as an interface, to avoid dependency of an implementation.
2. We give it a default value in the constructor, for the case DI system can't find a provider for it: a new class of the GraphlQLContextIDFactory

My main concern here is to avoid depending on such concrete classes and leverage the customization of such ContextIdFactory. Do you think it's feasible?

kamilmysliwiec:
Resolved in this PR https://github.com/nestjs/graphql/pull/2208

Issue #2231, opened 2022-06-08
Federation Directive on Entity class means fields are missing when using recent graphql dependencies. 


eddiesholl:
### Did you read the migration guide?

- [X] I have read the whole migration guide

### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Describe the regression

Starting point: I created a very vanilla nest js app, following the guides to set up the repo via the CLI, and then add dependencies needed for graphql federation. This includes getting graphql v16, for example.

When defining types via code-first, using an @Directive on a basic entity definition class means that the graphql parser can't find any fields on the type, and throws an error. Something has broken in the Apollo and graphql dependencies in a recent release, as this can be reproduced with the sample 31 federation example, if you upgrade to the latest dependencies.

A simple entity definition like this throws the error, and if you remove the Directive the fields are found correctly:
```
import { ObjectType, Field, ID, Directive } from '@nestjs/graphql';

@ObjectType()
@Directive('@key(fields: "id")')
export class OrgGroup {
  @Field((type) => ID)
  id: number;

  @Field()
  name: string;

  @Field((type) => [ID])
  members: number[];
}

GraphQLValidationFailed: The schema is not a valid GraphQL schema.. Caused by:
[<unnamed>] Type OrgGroup must define one or more fields.

    at new GraphQLErrorExt (/foo/node_modules/@apollo/core-schema/src/error.ts:23:5)
    at err (/foo/pathzero/pathzero.starter-kit/node_modules/@apollo/core-schema/src/error.ts:67:17)
    at ErrGraphQLValidationFailed (/foo/node_modules/@apollo/subgraph/node_modules/@apollo/federation-internals/src/definitions.ts:52:6)
    at addSubgraphToError (/foo/@apollo/subgraph/node_modules/@apollo/federation-internals/src/federation.ts:1506:84)
    at Subgraph.validate (/foo/node_modules/@apollo/subgraph/node_modules/@apollo/federation-internals/src/federation.ts:1389:15)


```

I've reproduced the issue using the nest sample 31, and upgrading the package.json dependencies to match what I've ended up with from running a nest initialisation today. See the branch here: https://github.com/eddiesholl/nest/tree/graphql-16-directive-bug

### Minimum reproduction code

Reproduced using the branch here:
https://github.com/eddiesholl/nest/tree/graphql-16-directive-bug

- Go to /sample/31-graphql-federation-code-first/posts-application
- `npm install`
- `npm run start:dev`

[...]

Issue #2141, opened 2022-04-19
Support for Apollo Federation 2


stringbeans:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Currently, it's unclear how to get Apollo Federation 2 to work with NestJS, both code-first and schema-first. I see two main points that need to be addressed:

1. adding the new `extend schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key", "@shareable"])` statement at the top of the typedefs
2. using the `buildSubgraphSchema` function from `@apollo/subgraph` to create the schema

(as referenced here: https://www.apollographql.com/docs/federation/subgraphs)

### Describe the solution you'd like

Perhaps a new driver (or replace the existing federation driver) to support the latest versions of:

```
"@apollo/gateway": "2.0.1",
"@apollo/subgraph": "2.0.1",
```

Internally, the driver would call `buildSubgraphSchema` function from `@apollo/subgraph` to create the schema as suggested in the official apollo federation documentation

### Teachability, documentation, adoption, migration strategy

The documentation can reference apollo's official federation docs to understand the differences between Federation 1 and Federation 2. 

Federation 2 is also a drop-in solution which is backwards compatible with Federation 1. Usage of new features in Federation 2 are opt-in and will require some minor adjustments to the typedefs

### What is the motivation / use case for changing the behavior?

Apollo Federation 2 simplifies the defining of subgraphs as well as how it handles merging and collisions. More here: https://www.apollographql.com/docs/federation/federation-2/new-in-federation-2

stringbeans:
@kamilmysliwiec i think there are two separate issues. that's why i had opened up the other issue here (https://github.com/nestjs/nest/issues/9472)

this current issue is about supporting Apollo Federation 2

the linked issue (https://github.com/nestjs/nest/issues/9472) is about supporting Federation 1 with the current dependencies

just wanted to mention that. by supporting federation 2 we kill two birds with one stone. but we could also treat it separately :+1:

stringbeans:
(regarding the support of Apollo Federation 2)

if there was a way to fetch the resolvers map, then i could do something like so...

```
GraphQLModule.forRoot<ApolloDriverConfig>({
      driver: ApolloDriver,
      autoSchemaFile: true,
      transformAutoSchemaFile: true,
      transformSchema: (schema) => {
        const extendSchema = gql`
[...]

PR #2251, opened 2022-06-16
Support Apollo Federation 2

greguintow:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #2141 

This also fixes:
- #2231

## What is the new behavior?

Allows to add a property called `useFed2` inside `autoSchemaFile` that will make the necessary schema changes to be considered a federation 2 schema:
https://github.com/greguintow/nestjs-graphql/blob/cb6090d766c78af96a628358c10d7efaca61ba9e/packages/apollo/tests/code-first-graphql-federation2/users-service/federation-users.module.ts#L12-L14

By default will add:
```graphql
extend schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key", "@shareable", "@external", "@override", "@requires"])
```
But that can be customized as `useFed2` type is  `boolean | Federation2Config`

- [ ] Yes
- [x] No

cc @trevor-scheer @hwillson


ChrisBates:
As a suggestion, get rid of `useFed2` and turn `autoSchemaFile` into `boolean | SchemaFileConfig`

Then `SchemaFileConfig` can be configurable to the role of the server - standard GQL, Federation Gateway, Subgraph, version, and whatever else might appear in the future.

Handle a `autoShemaFile: true` to default to the current standard functionality, and now there's a proper way to handle future changes.

kamilmysliwiec:
Thanks for your contribution @greguintow. I think @ChrisBates suggestion about using the `autoSchemaFile` makes sense & is more future-proof. Could you apply this change? 🤞

greguintow:
> Thanks for your contribution @greguintow. I think @ChrisBates suggestion about using the `autoSchemaFile` makes sense & is more future-proof. Could you apply this change? 🤞

Thanks @kamilmysliwiec and @ChrisBates , I'll make the changes shortly

greguintow:
@kamilmysliwiec @ChrisBates I've finished the changes, please review it!

packages/graphql/lib/interfaces/schema-file-config.interface.ts
```diff
@@ -0,0 +1,40 @@
...
+   * If enabled, it will use federation 2 schema
+   * 
+   * **Note:** You need to have installed @apollo/subgraph@^2.0.0 and enable `autoSchemaFile`
+   * 
+   * This will add to your schema:
+   * ```graphql
+      extend schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key", "@shareable", "@external", "@override", "@requires"])
+   * ```
+  */
+  useFed2?: UseFed2Value;
```
Teagan42, line 32: Not to be pedantic- but wouldn't `"federationVersion"?: FederationVersion` (an enum or similar) be more flexible to future changes?

greguintow: @Teagan42 I think that we wouldn't have such a big change as we had with apollo federation 2 vs 1 in any near future. Due to to how is setup, I think would be best to just be like that

ChrisBates: I agree with @Teagan42 - we don't know what the future brings, but we know for certain that Fed V2 won't be the latest version forever and we'll eventually need to deprecate a `useFed2` property. At least `federationVersion` can have the default updated to adopt the current version, and overwritten for backward compatibility (where people need it).

Teagan42: If this were GraphQL specification change, no way useFed2 would hold water. In that spirit, please accommodate future development with a scalable, flexible, extendible, and maintainable pattern

Teagan42, line 32: [Change Request] `federationConfig: FederationConfig`

packages/graphql/lib/utils/auto-schema-file.util.ts
```diff
@@ -0,0 +1,23 @@
+import { isString, isObject } from '@nestjs/common/utils/shared.utils';
+import { AutoSchemaFileValue, UseFed2Value } from '../interfaces';
+
+export function getPathForAutoSchemaFile(
```
Teagan42, line 4: [Change Request] Factory or Provider pattern

Teagan42 (changes requested):

packages/graphql/lib/interfaces/schema-file-config.interface.ts
kamilmysliwiec: @greguintow Let me know if you're interested in making these changes, otherwise, I'll take over! 🙌

greguintow: @kamilmysliwiec I'd have time for it this weekend, if that's okay

cristian-moreno-ruiz:
Can't get a federation working currently, are there any news on this PR? Which are the latest versions I can use to make a federated graph work with nestjs otherwise?

NeutronScott12:
> Can't get a federation working currently, are there any news on this PR? Which are the latest versions I can use to make a federated graph work with nestjs otherwise?

I use "@apollo/federation": "^0.36.1", "@nestjs/graphql": "^10.0.18", "graphql": "^15.7.2", you have to avoid apollo federation 2 packages and not go above 15.7 in graphql.

cristian-moreno-ruiz:
> > Can't get a federation working currently, are there any news on this PR? Which are the latest versions I can use to make a federated graph work with nestjs otherwise?
> 
> I use "@apollo/federation": "^0.36.1", "@nestjs/graphql": "^10.0.18", "graphql": "^15.7.2", you have to avoid apollo federation 2 packages and not go above 15.7 in graphql.

Turns out I was using `@apollo/subgraph: 2.0.5`, looks like this was the issue, thanks!

hwillson:
@NeutronScott12 @cristian-moreno-ruiz just to note, `@apollo/federation` is an older package and doesn't support Federation 2. It will work for now as along as you aren't using Federation 2 specific features. `@apollo/subgraph` is the new Federation 2 package which this PR will add support for.

ProductOfAmerica:
I'm so confused... is this working or not? Documentation says it should "just work" by using `federation: 2`, yet I'm getting this error: `ERROR GraphQLSchemaValidationError: Unknown directive "@link".`

### Using federation 2
webstorm64_dnEpJfnffH

### Not using federation 2
webstorm64_gM6skmZ5eb

What am I doing wrong?


My code:


#### app.module.ts
```typescript
    GraphQLModule.forRoot<ApolloFederationDriverConfig>({
      driver: ApolloFederationDriver,
      autoSchemaFile: {
        federation: 2,
        path: path.join(process.cwd(), 'packages/accounts/src/schema.gql'),
      },
      plugins: [ApolloServerPluginInlineTraceDisabled()],
      sortSchema: true,
      buildSchemaOptions: {
        orphanedTypes: [User],
      },
      transformAutoSchemaFile: true,
    }),
```

#### user.model.ts
```typescript
import { ObjectType, Field, ID, Directive } from '@nestjs/graphql';

@ObjectType()
@Directive('@key(fields: "id")')
export class User {
  @Field((type) => ID)
  id: number;
}
```

Issue #2331, opened 2022-08-15
Field description overwritten when using model extends and ResolveField()


mattleff:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When a field is defined in a base class that is extended by a model and that model has a resolver with a `ResolveField()` implemented for that base class field, the field `description` (and `deprecationReason`, etc.) in the base class is lost. This applies even if the `ResolveField()` in the resolver defines no options. 

### Minimum reproduction code

https://codesandbox.io/s/objective-ramanujan-53gmy7

### Steps to reproduce

1. Load CodeSandbox
2. Go to https://53gmy7.sse.codesandbox.io/graphql
3. Hover or check docs for `Cat.name`

### Expected behavior

`Cat.name` should have the description `"The name of the cat"` from `CatBase`. `Cat.isHidden` (which does not have a `ResolveField()` in the resolver) and `Cat.age` (which has a `ResolveField()` in the resolver, but the description is defined in the model instead of the base class) both correctly show their descriptions.

### Package version

10.0.21

### Graphql version

`graphql`: 16.5.0
`apollo-server-express`: 3.10.1

### NestJS version

9.0.9

### Node.js version

16.x

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [X] Linux

### Other

This is similar to #1012, #1424, and #2285. I'm happy to send a PR if you agree this is something that should be fixed.

kamilmysliwiec:
PRs are more than welcome @mattleff! 🙌

adalinesimonian:
This issue has not been resolved for resolvers which provide their property names via the decorator. For example:

```typescript
// The following will overwrite the base model options:
@ResolveField('fieldName', () => [Type])
async getFieldName(@Parent() parentType: ParentType) {
  // ...
}

// The following will not:
@ResolveField(() => [Type])
async fieldName(@Parent() parentType: ParentType) {
  // ...
}
```

PR #2337, opened 2022-08-16
fix(code-first): Use base model field options with ResolveField()

mattleff:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #2331

## What is the new behavior?

Field options from model base classes are used when a `ResolveField()` is defined for those fields.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/graphql/lib/schema-builder/storages/type-metadata.storage.ts
+++ b/packages/graphql/lib/schema-builder/storages/type-metadata.storage.ts
@@ -338,24 +338,8 @@ export class TypeMetadataStorageHost {
   }
 
   private compileExternalFieldResolverMetadata(item: FieldResolverMetadata) {
-    const objectTypeRef = this.metadataByTargetCollection
-      .get(item.target)
-      .resolver.typeFn();
-
-    const objectOrInterfaceTypeMetadata =
-      this.metadataByTargetCollection.get(objectTypeRef).objectType ||
-      this.metadataByTargetCollection.get(objectTypeRef).interface;
-
-    if (!objectOrInterfaceTypeMetadata) {
-      throw new CannotDetermineHostTypeError(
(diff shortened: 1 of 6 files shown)


packages/graphql/lib/schema-builder/storages/type-metadata.storage.ts
```diff
@@ -394,6 +378,41 @@ export class TypeMetadataStorageHost {
...
+      .resolver.typeFn();
+    const getTypeMetadata = (objectTypeRef: any) =>
+      this.metadataByTargetCollection.get(objectTypeRef).objectType ||
+      this.metadataByTargetCollection.get(objectTypeRef).interface;
+    for (
+      let _objectTypeRef = objectTypeRef;
+      _objectTypeRef?.prototype;
+      _objectTypeRef = Object.getPrototypeOf(_objectTypeRef)
+    ) {
+      const possibleTypeMetadata = getTypeMetadata(_objectTypeRef);
```
kamilmysliwiec, line 395: What if `IngredientBase` wasn't annotated with the `ObjectType` decorator? Would this method throw an exception? Is that a breaking change, or was that the case before too?

mattleff: @kamilmysliwiec Good point. Prior to this PR it wouldn't add the field without the decorator but it also wouldn't throw as long as the field was defined in the `Resolver`. I've added a case to ensure we don't throw when a) the field is defined in a base class that doesn't have the `ObjectType` decorator and b) there is a `Field()` defined for that field in the resolver.

kamilmysliwiec:
LGTM

PR #2343, opened 2022-08-17
fix(): Counting the number of arguments handle more complex cases

ln-e:
Before `getNumberOfArguments(ContextIdFactory.getByRequest)` was equals to 3. Now it is handle much more complex cases.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/graphql/lib/utils/get-number-of-arguments.util.ts
+++ b/packages/graphql/lib/utils/get-number-of-arguments.util.ts
@@ -7,26 +7,78 @@
  */
 export function getNumberOfArguments(fn: Function): number | undefined {
   // Removing newlines is necessary to use easier regex and handle multi-line functions
-  const functionAsStringWithouNewLines = fn.toString().replace(/\n/g, '');
+  const functionAsStringWithoutNewLines = fn.toString().replace(/\n/g, '');
 
-  const anythingEnclosedInParenthesesRegex = /\(.+\)/;
-
-  const regexMatchedArray = functionAsStringWithouNewLines.match(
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
What's the use case?

ln-e:
https://github.com/nestjs/graphql/blob/846ba4e2254fd8faeeb2d3d23e13e51901132e46/packages/graphql/lib/services/resolvers-explorer.service.ts#L48-L50

It is return 3 now.

kamilmysliwiec:
Again - what's the (real world) use case?

ln-e:
Again: you write a some ["backward compatibility level"](https://github.com/nestjs/graphql/blob/846ba4e2254fd8faeeb2d3d23e13e51901132e46/packages/graphql/lib/services/resolvers-explorer.service.ts#L342-L380) which is depends on incorrect result of buggy function. Although the result of the function is correct, the code inside does not work as expected. Is it ok you think?

kamilmysliwiec:
The goal was to make this function as fast as it could be without introducing breaking changes. We want to remove it eventually anyway. As long as the functionality work as expected, there's no reason to update it

ln-e:
> as fast as it could be

Well, the current implementation is heavily rely on Regex. Thats why the new one [is faster](https://jsbench.me/tpl78p7ev7/1) btw
Снимок экрана 2022-08-25 в 10 24 05

PR #2402, opened 2022-09-14
Various performance boosts in nests/graphql

roypeled:
## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When bootstrapping a gql application, having many schemas results in a very large memory footprint.
For our project with 200 schemas, we couldn't run the server because it consumed past 8GB memory and died on OOM.

Issue Number: N/A

## What is the new behavior?
Reducing the amount of redundant compilations of the GQL data object model (metadata wise) is now possible because for the switch to map based metadata storage. Removing the redundant metadata assignments produces a tremendous reduction of memory usage (from 8GB+ to 700MB) 

- [ ] Yes
- [x] No


rezoled:
Hey @kamilmysliwiec, I've made some improvements that drastically reduce the memory footprint of a GQL server. Could you please verify and merge this?

roypeled:
Hey @kamilmysliwiec, I've made some improvements that drastically reduce the memory footprint of a GQL server. Could you please verify and merge this?

roypeled:
1) getObjectOrInterfaceTypeNameIfExists - This function was not fixed on the move to map based metadata storage. Now instead of looking inside an entire metadata array, it fetches the item from the map

2) MetadataCollectionModel - I noticed that the interfaces list recieves duplicate instances of the same interface, so I switched the implementation to a map to remvoe duplicates.

3) LazyMetadataStorageHost.load - In very large applications (ie several millions of metadata object) we encountered an invalid arugments size thrown from V8, this is because the spread operator behind the scenes translates the values to multiple arguments. Switching to flat() solves this issue.

4) GraphQLSchemaFactory.create & LazyMetadataStorageHost.updateStorage - We noticed that the entire GQL metadata model was being rebuilt several times during the bootstrap of the application, I assume this was done because of the need to clear and rebuild the arrays every time. Now when we use maps we can't get duplicates in the metadata store, so I verify that the updateStorage callback is called only once, and I removed the clear call from the GraphQLSchemaFactory.create(). This results in a tremendous memory consumption boost! (in our case, from 8GB to 700MB)

packages/graphql/lib/schema-builder/graphql-schema.factory.ts
```diff
@@ -60,7 +60,6 @@ export class GraphQLSchemaFactory {
       options = scalarsOrOptions;
     }
 
-    TypeMetadataStorage.clear();
```
kamilmysliwiec, line 63: Why this line is no longer required?

rezoled: Because we save the metadata in maps. 
What I think happened was that when we used array to save the metadata, and called updateStorage in lazy-metadata.storage multiple times, the metadata would get duplicated in the storage.

For performance optimization, I removed redundant updateStorage calls, but now calling GraphQLSchemaFactory.create called for TypeMetadataStorage.clear (it expected the the metadata would be collected again via updateStorage), and the metadata would not get re-collected (because I removed duplicate updateStorage calls).

So I removed the clear call. 

Why this works?
1) Since TypeMetadataStorage has changed to maps, there is no chance of getting duplicate metadata values. 
2) When we have no duplicates, there is no need to reset and re-collect the metadata.

I hope my explanation is clear.

rezoled: The only thing that bothers me, is that TypeMetadataStorage is a singleton that is generated on app bootstrap, and is not bounded by a NestJS instance

packages/graphql/lib/decorators/resolver.decorator.ts
```diff
@@ -20,10 +20,8 @@ function getObjectOrInterfaceTypeNameIfExists(
...
   const ctor = getClassOrUndefined(nameOrType);
-  const objectTypesMetadata = TypeMetadataStorage.getObjectTypesMetadata();
-  const objectMetadata = objectTypesMetadata.find(
-    (type) => type.target === ctor,
-  );
+  const objectMetadata =
+    TypeMetadataStorage.getObjectTypeMetadataByTarget(ctor);
   if (!objectMetadata) {
     const interfaceTypesMetadata = TypeMetadataStorage.getInterfacesMetadata();
     const interfaceMetadata = interfaceTypesMetadata.find(
```
kamilmysliwiec, line 27: Shouldn't we use `getInterfaceMetadataByTarget()` here as well?

packages/graphql/lib/schema-builder/collections/metadata-by-target.collection.ts
```diff
@@ -1,10 +1,11 @@
...
+import { ClassMetadata } from '../metadata';
 
 export class MetadataByTargetCollection {
   public readonly all: MetadataCollectionModel = {
     argumentType: [],
-    interface: [],
+    interface: new Map<Function, ClassMetadata>(),
     inputType: [],
     objectType: [],
     resolver: [],
```
kamilmysliwiec, line 11: Do you think we should refactor other collections too? (especially objectType?)

packages/graphql/lib/decorators/resolver.decorator.ts
rezoled: Definitley. Fixing

packages/graphql/lib/schema-builder/collections/metadata-by-target.collection.ts
rezoled: I agree, but I'm afraid of introducing a lot of changes. Some tests rely on a specific order (arguments for example), moving it to map would break the order.
I think that maybe there's a reason to rethink how the metadata storage works? Maybe a complete overhaul to have it much more optimized?

kamilmysliwiec:
image

Build is broken now @roypeled

Issue #1907, opened 2021-12-03
Resolver metadata cache returns wrong results


jakubnavratil:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When using two equally named resolvers even when they are in diffrent modules, and both have equally named methods, Args decorator doesnt propagate data into the last one.
```typescript
@Resolver()
export class UserResolver {
  @Mutation(() => String, { name: 'moduleALogin' })
  login(@Args('code') code: string) {
    return code;
  }
}

@Resolver()
export class UserResolver {
  @Mutation(() => String, { name: 'moduleBLogin' })
  login(@Args('username') username: string) {
    return username; // BUG: undefined
  }
}
```

### Minimum reproduction code

https://github.com/jakubnavratil/nestjs-resolvers-bug

### Steps to reproduce

npm i
npm run start
open http://localhost:3000/graphql

run both, B fails
```graphql
mutation A {
  moduleALogin(code:"code")
}
mutation B {
  moduleBLogin(username:"username")
}
```

### Expected behavior

Works without need to make unique methods name across multiple modules.

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [X] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### Packages versions

 "@nestjs/common": "^8.0.0",
"@nestjs/core": "^8.0.0",
"@nestjs/graphql": "^9.1.2",

### In which operating systems have you tested?

- [ ] macOS
- [X] Windows
- [ ] Linux

### Other

It comes down to this commit https://github.com/nestjs/nest/commit/035348e07d22897b83b1ff20dcc6ea24094c49c3
which fixes this https://github.com/nestjs/nest/issues/3231 and introduced uniquely generated ID for controllers but doesnt do that for Resolvers (providers)

This ID is then used here https://github.com/nestjs/nest/blob/master/packages/core/helpers/handler-metadata-storage.ts#L62 
but if it is missing, it uses class name to generate keys. 
Hence both `login` methods in my example has same keys.

[...]

PR #2408, opened 2022-09-19
fix(@nestjs/graphql): set unique ids on resolvers to fix cache errors

koyedele:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The resolver metadata cache currently uses the resolver name for the key as shown below

https://github.com/nestjs/nest/blob/39ec4cda7363aac3551ff0f1b83feb41ba2f0bf6/packages/core/helpers/handler-metadata-storage.ts#L60-L63

This causes issues if we have two resolvers in different modules but with the same name and identical method signatures. For controllers, each controller gets a unique ID that is used as the key in the cache as shown below:

https://github.com/nestjs/nest/blob/39ec4cda7363aac3551ff0f1b83feb41ba2f0bf6/packages/core/injector/module.ts#L461-L462

Add code to set up unique IDs for resolvers as well so that cache retrieval is correct.

Issue Number: #1907

## What is the new behavior?

Each resolver gets a unique ID so that it can be retrieved correctly from the metadata cache. The solution re-uses the ID key that controllers also use so I worry that that may cause confusion. 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/apollo/tests/code-first-duplicate-resolvers/app.module.ts
+++ b/packages/apollo/tests/code-first-duplicate-resolvers/app.module.ts
@@ -1,0 +1,21 @@
+import { ApolloDriver } from '@nestjs/apollo';
+import { Module } from '@nestjs/common';
+import { GraphQLModule } from '@nestjs/graphql';
+import { ModuleAModule } from './module-a/module-a.module';
+import { ModuleBModule } from './module-b/module-b.module';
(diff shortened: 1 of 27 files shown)


kamilmysliwiec:
LGTM

Issue #2255, opened 2022-06-19
Unused `ObjectType` referencing custom scalar results in error during schema generation


SimonCockx:
### Current behavior

When I define a custom scalar (e.g., `@Scalar('MyCustomScalar')`) which I reference in an `ObjectType` that isn't exposed through any resolver, I run into an error `Error: "MyCustomScalar" defined in resolvers, but not in schema` during schema generation.

### Minimum reproduction code

https://github.com/SimonCockx/nest-gql-scalar-bug

### Steps to reproduce

1. Clone `https://github.com/SimonCockx/nest-gql-scalar-bug`
2. `npm install`
3. `nest start`
See error.

### Expected behavior

I would expect one of these two:
1. `@nestjs/graphql` successfully determines that the custom scalar is not necessary, so it does not generate it in the schema, but it also doesn't produce an error (which is the case now).
2. `@nestjs/graphql` just generates the scalar without detecting whether it is referenced in some way. (probably simpler, but I saw the current code already does some reference detection to find only necessary types, so option (1) might be better)

### Package version

10.0.16

### Graphql version

`graphql`: `16.5.0`
`apollo-server-express`: `3.9.0`

### NestJS version

8.0.0

### Node.js version

16.15.1

### In which operating systems have you tested?

- [ ] macOS
- [X] Windows
- [ ] Linux

iamkhalidbashir:
Same issue for me
It fails on tests but runs fine on dev server

tuxmachine:
I can confirm, it still occurs in latest. I've opened a PR with a potential fix.

kamilmysliwiec:
Let's track this here https://github.com/nestjs/graphql/pull/2428

PR #2428, opened 2022-10-01
fix(@nestjs/graphql): always add scalars to generated schema

tuxmachine:
`makeExecutableSchema` throws if a resolver is defined without being present in the schema. We should include scalars even if they're not used by any field (yet)

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

If an app uses GraphQL code-first schema generation and had a scalar without referencing this from any resolver, the app will throw an error upon startup: `Error: "XYZ" defined in resolvers, but not in schema`

Issue Number: #2255 

## What is the new behavior?

Always include scalars in the generated schema by explicitly adding them to the list of types, instead of relying on the implicit inclusion via a resolver in- or output.

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/packages/apollo/tests/code-first/other/sample-scalar.ts
+++ b/packages/apollo/tests/code-first/other/sample-scalar.ts
@@ -1,0 +1,30 @@
+import { CustomScalar, Scalar } from '@nestjs/graphql';
+import { Kind } from 'graphql';
+
+@Scalar('SampleScalar')
+export class SampleScalar implements CustomScalar<string, string> {
+  description = 'A sample scalar';
+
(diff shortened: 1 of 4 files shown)


jmcdo29 (approved):
Looks good to me. I'll let Kamil say if there's anything that seems off about adding the scalars to the types

kamilmysliwiec:
LGTM 🙌 

I'll merge it in shortly

GerardEmmen:
We're having the same issue. Would be really much helped with this fix!

PR #2500, opened 2022-11-09
fix: enable non-nullable, defaulted fields in code-first schema

sam-super:
## PR Checklist

## PR Type

- [x] Bugfix

When a not-nullable field has a defaultValue it generates a nullable input field:
```typescript
@InputType()
export class AcmeInput {
  @Field({ nullable: false, defaultValue: 'foo' })
  bar: string;
}
```
it generates:
```gql
input AcmeInput {
  bar: String = "foo"
}
```

## What is the new behavior?
The above will generate this schema:
```gql
input AcmeInput {
  bar: String! = "foo"
}
```

- [x] No

## Other information

Some of the tests (e.g. the snapshots) capture the erroneous functionality, so to keep this PR specific to the change, some existing fields had `nullable: true` set to match the existing tests.

### DIFF
--- a/packages/graphql/lib/schema-builder/services/type-mapper.service.ts
+++ b/packages/graphql/lib/schema-builder/services/type-mapper.service.ts
@@ -67,18 +67,7 @@ export class TypeMapperSevice {
       );
     }
 
-    let isNotNullable: boolean;
-    if (isInputTypeCtx) {
-      /**
-       * The input values (e.g., args) remain "nullable"
-       * even if the "defaultValue" is specified.
-       */
-      isNotNullable =
-        isUndefined(options.defaultValue) &&
-        (!options.nullable || options.nullable === 'items');
-    } else {
-      isNotNullable = !options.nullable || options.nullable === 'items';
-    }
+    const isNotNullable = !options.nullable || options.nullable === 'items';
     return isNotNullable
       ? (new GraphQLNonNull(graphqlType) as T)
       : (graphqlType as T);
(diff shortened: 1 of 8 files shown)


packages/graphql/lib/schema-builder/services/type-mapper.service.ts
```diff
@@ -67,18 +67,7 @@ export class TypeMapperSevice {
       );
     }
 
-    let isNotNullable: boolean;
-    if (isInputTypeCtx) {
```
kamilmysliwiec, line 71: Since we no longer use the "isInputTypeCtx" argument, let's just remove it from the method signature.

sam-super: Good point. Have removed.

kamilmysliwiec:
lgtm

piotr-pawlowski:
This "fix" doesn't make sense to me. Now Apollo Kotlin doesn't allow to omit params with default values. Before this fix I could write `foo = Input.absent()` and I knew that the default value would be applied. Now I have to write `foo = "bar"`, so my app needs to know what is default value for given param. 
The second thing is that you wrote: "Does this PR introduce a breaking change? No". It's not true because from now on parameters with default values are non-nullable. So graphql-inspector marks these fields as breaking changes.
To keep backward compability I am required to add explicitly `{ nullable: true }` to each field.

zq0904:
I was using "@nestjs/graphql": "^10.1.3 ", but the online version updates automatically to 10.2.0

I have a large number of similar inputtypes

For example:
@InputType()
export class ProjectsOptPagingInput extends ProjectOptPagingInput {
@Field(() => Int)
projectId: Project['id'] = 0;
}

This change is a destructive change. I would like to know why it was modified. Is the same with version 11?

wodka:
How can this be in a patch release? it completely breaks the old behaviour of defaultValue! Now per default all are mandatory, whereas they have been optional!

-> this should be mentioned as a breaking change and should have been released in v11!

PR #2559, opened 2022-12-16
fix(graphql): similar directive sdls on multiple fields fail

harm-less:
Similar directive sdls on multiple fields fail if applied to the same target. I stumbled upon this while migrating from NestJS 8 to 9. It still baffles me how this could have happened from all the code I analysed and read, but I guess there's a good underlying reason for this bug.

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

I ran into a problem in which the following code doesn't not register all applied directives:

```ts
@Resolver(() => TypeSomething)
export class TwoResolversWithSimilarDirectives {

	@Query((returns) => TypeSomething)
	@Directive('@hasPermission(permissions: APP_ACCOUNTS_READ)')
	@Directive('@limit')
	async resolver1() {
		return {}
	}

	@Query((returns) => TypeSomething)
	@Directive('@hasPermission(permissions: APP_ACCOUNTS_READ)')
	@Directive('@limit')
	async resolver2() {
		return {}
	}
}
```

As can be seen, I've got 2 resolver methods in 1 target both of which the same SDLs are applied to. Yet, what currently happens is that the second resolver only has one directive applied to the GraphQL schema and not both as intended. 

Issue Number: N/A

## What is the new behavior?
All expected directives are applied to the corresponding fields

- [ ] Yes
- [x] No

### DIFF
--- a/packages/graphql/lib/schema-builder/collections/field-directive.collection.ts
+++ b/packages/graphql/lib/schema-builder/collections/field-directive.collection.ts
@@ -4,9 +4,11 @@ import { MetadataListByNameCollection } from './metadata-list-by-name.collection
 export class FieldDirectiveCollection extends MetadataListByNameCollection<PropertyDirectiveMetadata> {
   sdls = new Set<string>();
   fieldNames = new Set<string>();
+  uniqueCombinations = new Set<string>();
 
   add(value: PropertyDirectiveMetadata) {
-    if (this.sdls.has(value.sdl) && this.fieldNames.has(value.fieldName)) {
+    const combinationKey = `${value.sdl}${value.fieldName}`;
+    if (this.uniqueCombinations.has(combinationKey)) {
       return;
     }
 
@@ -14,6 +16,7 @@ export class FieldDirectiveCollection extends MetadataListByNameCollection<Prope
 
     this.sdls.add(value.sdl);
     this.fieldNames.add(value.fieldName);
+    this.uniqueCombinations.add(combinationKey);
     this.globalArray?.push(value);
   }
 }
--- a/packages/graphql/lib/schema-builder/collections/field-directive.collection.ts
+++ b/packages/graphql/lib/schema-builder/collections/field-directive.collection.ts
@@ -8,7 +8,9 @@ export class FieldDirectiveCollection extends MetadataListByNameCollection<Prope
 
   add(value: PropertyDirectiveMetadata) {
     const combinationKey = `${value.sdl}${value.fieldName}`;
-    if (this.uniqueCombinations.has(combinationKey))   return;
+    if (this.uniqueCombinations.has(combinationKey)) { 
+        return;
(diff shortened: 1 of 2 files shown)


packages/graphql/lib/schema-builder/collections/field-directive.collection.ts
```diff
@@ -4,16 +4,19 @@ import { MetadataListByNameCollection } from './metadata-list-by-name.collection
...
   sdls = new Set<string>();
   fieldNames = new Set<string>();
+  uniqueCombinations = new Set<string>();
 
   add(value: PropertyDirectiveMetadata) {
-    if (this.sdls.has(value.sdl) && this.fieldNames.has(value.fieldName)) {
+    const combinationKey = `${value.sdl}${value.fieldName}`;
+    if (this.uniqueCombinations.has(combinationKey)) {
       return;
     }
```
ProKashif, line 13:
```suggestion
    if (this.uniqueCombinations.has(combinationKey))   return;
```

ProKashif (changes requested):

packages/graphql/lib/schema-builder/collections/field-directive.collection.ts
```diff
@@ -4,16 +4,17 @@ import { MetadataListByNameCollection } from './metadata-list-by-name.collection
...
   sdls = new Set<string>();
   fieldNames = new Set<string>();
+  uniqueCombinations = new Set<string>();
 
   add(value: PropertyDirectiveMetadata) {
-    if (this.sdls.has(value.sdl) && this.fieldNames.has(value.fieldName)) {
-      return;
-    }
+    const combinationKey = `${value.sdl}${value.fieldName}`;
+    if (this.uniqueCombinations.has(combinationKey))   return;
```
kamilmysliwiec, line 11:
```suggestion
    if (this.uniqueCombinations.has(combinationKey)) { 
        return;
    }
```

packages/graphql/lib/schema-builder/collections/field-directive.collection.ts
```diff
@@ -4,16 +4,19 @@ import { MetadataListByNameCollection } from './metadata-list-by-name.collection
...
   sdls = new Set<string>();
   fieldNames = new Set<string>();
+  uniqueCombinations = new Set<string>();
 
   add(value: PropertyDirectiveMetadata) {
-    if (this.sdls.has(value.sdl) && this.fieldNames.has(value.fieldName)) {
-      return;
+    const combinationKey = `${value.sdl}${value.fieldName}`;
+    if (this.uniqueCombinations.has(combinationKey)) { 
+        return;
```
kamilmysliwiec, line 12:
```suggestion
      return;
```

kamilmysliwiec:
LGTM

Issue #2360, opened 2022-08-26
Regression in 10.0.19: Error: Cannot determine a GraphQL input type ("....") for the "...". Make sure your class is decorated with an appropriate decorator.     at InputTypeFactory.create


andreialecu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

After upgrading to 10.0.22 I started getting a weird error about in a pretty big project:

```
Error: Cannot determine a GraphQL input type ("ClubModel") for the "club". Make sure your class is decorated with an appropriate decorator.
    at InputTypeFactory.create
```

There was no context as to where this error originated, so I had to add various `console.log`s inside `@nestjs/graphql` to find it.

I eventually tracked it down to this piece of code:

![image](https://user-images.githubusercontent.com/697707/186908400-72b544c4-b9c0-459e-843b-5ec0de040b59.png)

The error was about an **InputType** being expected on a `@ResolveField()` so this is very bizarre.

Here's the relevant code:

```ts
const isAbstract = getAppCaller() === AppCaller.PublicApi ? true : false;

function resolveType(value: { kind: typeof ALL_CLOCK_UNIT_KINDS[number] }) {
  return ALL_CLOCK_UNIT_MODELS.find((c) => c.kind === value.kind);
}

@InterfaceType("ClockDeviceUnit", {
  isAbstract,
  resolveType,
})
@InputType("ClockDeviceUnitBaseInput", { isAbstract })
export class ClockDeviceUnitBaseModel {
  @prop({ type: () => String, required: true })
  kind!: ClockDeviceUnitKinds;

  @prop({ required: true })
  @Field()
  clubId!: string;
}

@Resolver(() => ClockDeviceUnitBaseModel)
export class ClockDeviceUnitResolver {
  constructor(private readonly clubService: ClubsService) {}

  @ResolveField(() => ClubModel)
  async club(@Parent() unit: ClockDeviceUnitBaseModel): Promise<ClubModel> {
    return this.clubService.getClub(unit.clubId);
  }
}
```

### Minimum reproduction code

nope

### Expected behavior

A `@ResolveField()` should not participate in inputs, I suspect a refactoring broke this.

### Package version

10.0.19

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [ ] Linux

JonnyBGod:
I have the same issue just for reference:

```ts
  @ResolveField(() => Organization)
[...]

PR #2724, opened 2023-03-19
fix: metadata storage refactor caused a bug in field resolver

roypeled:
When creating a field resolver with an input we reached a bad edge case where the fields array was fetched via reference instead of an immutable array. This caused artifacts appearing in the array and bad metadata appeared when trying to parse the schema.

Returning a new array from the metadata instead of the reference for the existing array solves this issue

## PR Checklist

## PR Type

- [x] Bugfix

https://github.com/nestjs/graphql/issues/2360#issuecomment-1475362011

Issue Number: 2360

## What is the new behavior?
Fields resolver works correctly.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/graphql/lib/schema-builder/collections/metadata-by-name.collection.ts
+++ b/packages/graphql/lib/schema-builder/collections/metadata-by-name.collection.ts
@@ -3,7 +3,7 @@ export class MetadataByNameCollection<T> {
   protected all: (T extends any[] ? T[number] : T)[] = [];
 
   getAll() {
-    return this.all;
+    return [...this.all];
   }
 
   getByName(name: string) {


kamilmysliwiec:
lgtm

jhpierce:
Thank you both for the fix. Is there any chance this could be backported to v10?

roypeled:
I don’t see a reason not to, it’s a small fix. 
@kamilmysliwiec should I create a pull request?

jhpierce:
@kamilmysliwiec @roypeled I created [a PR](https://github.com/nestjs/graphql/pull/2739) to backport this fix to v10. Please advise on how to proceed, thank you!

Issue #2661, opened 2023-02-19
CLI plugin breaks paths when used with webpack alias (windows)


enricobuehler:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

The plugin breaks import paths when used in combination with webpack aliases. 
It seems to happen because webpack transforms the import paths to absolute paths, which start with the drive letter, e.g. `D:\...` and then the plugin suffixes this path with `./`. If really needed i can provide a minimal reproduction but since i know which code is responsible for the bug i thought it wouldn't help.

### Minimum reproduction code

-

### Steps to reproduce

1. Create nestjs project
2. Setup plugin and webpack with alias imports
3. Run the dev server
-> Import fails only for the files modified by the cli

### Expected behavior

Obviously the paths shouldn't break, however im not sure how to achieve this. In my case just removing the `./` suffix fixed it, could also provide a PR for this, but don't know about possible side effects. The suffix is added here: https://github.com/nestjs/graphql/blob/master/packages/graphql/lib/plugin/utils/plugin-utils.ts#L112

### Package version

10.1.6

### Graphql version

`graphql`: 16.6.0
`apollo-server-express`: 3.11.1

### NestJS version

9.2.1

### Node.js version

19.6.0

### In which operating systems have you tested?

- [X] macOS
- [X] Windows
- [ ] Linux

kamilmysliwiec:
We need a reproduction repository to fix this issue

enricobuehler:
Sure, here you go: https://github.com/enricobuehler/nestjs-graphql-paths-bug

enricobuehler:
Just tested it on macOS, there the issue doesn't occur.

all2pie:
@enricobuehler yes please it would be appreciated. 

You can also look into #2734  it might be related.

mikejpeters:
I'm running into the same issue on Windows (but it works fine on Linux).

The linked pull request fixes the compilation errors on Windows but then breaks things on Linux:

```
ERROR in ***/example.input.ts 6:23-47
Module not found: Error: Can't resolve 'other.input' in '/home/***/example/dto'
Did you mean './other.input'?
```

I tried to figure out what's going on, it seems like `relativePath` is actually an absolute path on windows:

```ts
  let relativePath = posix.relative(posix.dirname(fileName), importPath);
  relativePath = relativePath[0] !== '.' ? './' + relativePath : relativePath;
```

When this code runs on Windows for example:

[...]

PR #2788, opened 2023-04-23
fix(plugin): don't prefix absolute win paths

mikejpeters:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Compilation fails on Windows in certain cases when using the GraphQL plugin because absolute paths are prefixed with `./`.

Issue Number: #2661

## What is the new behavior?

Don't prefix absolute paths on Windows with `./`.

- [ ] Yes
- [x] No

## Other information

I think this is a temporary solution at best. I'm not sure the `posix` version of `path` methods should be being used for all the logic in the plugin, since things like `posix.relative()` do not behave properly with windows paths. I could maybe create another pull request to switch to the non-posix path methods, but not sure if I'm missing an important reason it's in use.

I also wanted to add a test but it's tricky:

```ts
describe('plugin-utils', () => {
  describe('replaceImportPath', () => {
  
    it('should not replace path on windows', () => {
      const actual = replaceImportPath(
          'import("C:\\\\project\\\\src\\\\author.model").Author',
          'C:\\project\\src\\post.model.ts',
      );

      expect(actual).toStrictEqual({
        typeReference: '"require("C:\\\\project\\\\src\\\\author.model").Author',
        importPath: 'C:/project/src/author.model',
      });
    });
```    

This test fails if run on Linux, since `path.isAbsolute('C:/...')` only returns `true` when run on Windows. I didn't want to complicate things by using Jest to mock path functions. Something like [jest-os-detection](https://github.com/doctolib/jest-os-detection) allows specifying whether a test should be skipped/run depending on the OS, but this is only very useful if CI also runs the test suite on Windows, and I didn't want to make too many changes for this pull request.

### DIFF
--- a/packages/graphql/lib/plugin/utils/plugin-utils.ts
+++ b/packages/graphql/lib/plugin/utils/plugin-utils.ts
@@ -1,5 +1,5 @@
 import { head } from 'lodash';
-import { posix } from 'path';
+import { isAbsolute, posix } from 'path';
 import * as ts from 'typescript';
 import {
   getText,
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
LGTM

PR #3211, opened 2024-06-07
fix(graphql): readonly arrays not allowed in decorator input

CarsonF:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

```ts

const interfaces = [A, B] as const;

@ObjectType({
  implements: interfaces, // error
})
```

- [ ] Yes
- [x] No

### DIFF
--- a/packages/graphql/lib/decorators/resolve-field.decorator.ts
+++ b/packages/graphql/lib/decorators/resolve-field.decorator.ts
@@ -39,7 +39,7 @@ export type ResolveFieldOptions<T = any> = BaseTypeOptions<T> & {
   /**
    * Array of middleware to apply.
    */
-  middleware?: FieldMiddleware[];
+  middleware?: readonly FieldMiddleware[];
 };
 
 /**
--- a/packages/graphql/lib/utils/types.ts
+++ b/packages/graphql/lib/utils/types.ts
@@ -1,0 +1,2 @@
+export type Many<T> = T | readonly T[];
+export type Lazy<T> = T | (() => T);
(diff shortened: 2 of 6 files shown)


packages/graphql/lib/utils/types.ts
```diff
@@ -0,0 +1,2 @@
+export type Many<T> = T | readonly T[];
```
micalevisk, line 1:
```suggestion
export type Many<T> = T | ReadonlyArray<T>;
```

this should work. See https://github.com/nestjs/nest/pull/13655

packages/graphql/lib/decorators/resolve-field.decorator.ts
```diff
@@ -39,7 +39,7 @@ export type ResolveFieldOptions<T = any> = BaseTypeOptions<T> & {
   /**
    * Array of middleware to apply.
    */
-  middleware?: FieldMiddleware[];
+  middleware?: readonly FieldMiddleware[];
```
micalevisk, line 42:
```suggestion
  middleware?: ReadonlyArray<FieldMiddleware>;
```

micalevisk (changes requested):
minor changes to support both array and read only arrays

CarsonF:
> minor changes to support both array and read only arrays

These represent the same thing. Mutable arrays are a superset of read-only ones. But if this style is preferred I can switch it.

micalevisk:
I just noticed something that would break due to that `ReadonlyArray`:

```ts
const imports: ResolveFieldOptions['middleware'] = [];
imports.push(() => {})
```

because we can't invoke `.push` on readonly arrays

and I believe that we shouldn't prevent that usage

PR #3433, opened 2025-01-09
feat(@nestjs/graphql): Adding  option

dustinsmith1024:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

This allows the consumers to not include the Apollo Drain Plugin.

- [ ] Yes
- [X] No

## Other information

I don't think this is breaking and I am pretty sure the default is set to keep the plugins. Working on testing this in my env. There really isn't unit tests for this code currently, if they are desired I can add them if you give me a little direction.

Thanks for considering!

### DIFF
--- a/packages/apollo/lib/drivers/apollo-base.driver.ts
+++ b/packages/apollo/lib/drivers/apollo-base.driver.ts
@@ -37,6 +37,10 @@ export abstract class ApolloBaseDriver<
     const platformName = httpAdapter.getType();
 
     if (platformName === 'express') {
+      console.log(
+        'ApolloBaseDriver.start() - platformName === express',
+        apolloOptions,
+      );
       await this.registerExpress(apolloOptions);
     } else if (platformName === 'fastify') {
       await this.registerFastify(apolloOptions);
@@ -50,10 +54,12 @@ export abstract class ApolloBaseDriver<
   }
 
   public async mergeDefaultOptions(options: T): Promise<T> {
+    console.log('passed in options', options);
     let defaults: ApolloDriverConfig = {
       path: '/graphql',
       fieldResolverEnhancers: [],
       stopOnTerminationSignals: false,
+      enableDrainPlugin: options.enableDrainPlugin ?? true,
     };
 
     if (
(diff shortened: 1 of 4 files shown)


kamilmysliwiec:
> This allows the consumers to not include the Apollo Drain Plugin.

What's the use case?

dustinsmith1024:
I am not sure this change is correct anymore but I'll walk through my testing.

We're having issues keeping the GraphQL server running after a SIGTERM is received. We're trying to user testing during a Kubernetes rolling reboot and we lose quite a few requests as the pods go down. 

We receive this error during app shutdown:

```
A GraphQL operation was received during server shutdown. The operation will fail. Consider draining the HTTP server on shutdown; see https://go.apollo.dev/s/drain for details.
```

This lead me to the drain plugin code. Originally I thought the Apollo `stopOnTerminationSignals: false` would help but it had no affect. I implemented my own Drain Plugin and that would run but server was already shut down. So I assumed some bug in that code and just wanted to disable it for now.

After I pushed this PR change to our environment I had to tweak it and turn off the `stop` in the GraphQL module onModuleDestroy hook.

https://github.com/nestjs/graphql/blob/f4c9a42154e4f31fa2573815b000b80c2c0fb015/packages/graphql/lib/graphql.module.ts#L76

```
async onModuleDestroy() {
  await this._graphQlAdapter.stop();
}
```

So my current understanding is that this graphql adapter stop method is shutting down the GraphQL server and we have no control over cleanup or keeping the server alive for a few seconds like we can do with normal lifecycle hooks.

If there is a way to override this method than I could just do that. Haven't tried that yet though.

It's possible I am missing something about how the hooks are all ordered. Started a thread in Discord as well last week: https://discord.com/channels/520622812742811698/1326963494821695579/1326963494821695579.

Thanks!

dustinsmith1024:
I can close this and make an issue instead?

dustinsmith1024:
This is the patch that is more helpful. I will close this out and work to see if there is a way to configure a wait time here.

```
diff --git a/dist/graphql.module.js b/dist/graphql.module.js
index 819a89302f9000bc4f15fd2330f97b95b6c9f6d7..b697abca60afefebe412916e6ae7866915c0d133 100644
--- a/dist/graphql.module.js
+++ b/dist/graphql.module.js
@@ -34,7 +34,8 @@ let GraphQLModule = GraphQLModule_1 = class GraphQLModule {
         this.metadataLoader = new metadata_loader_1.MetadataLoader();
     }
     async onModuleDestroy() {
-        await this._graphQlAdapter.stop();
+        console.log('Not stopping graphql server because pnpm patch override');
+        // await this._graphQlAdapter.stop();
     }
     static forRoot(options = {}) {
         this.assertDriver(options);
```

PR #3779, opened 2026-01-02
fix(graphql,apollo): ensures extensions are added to resolved fields

### LINKED ISSUE #3524: Code-First: Multiple GraphQL Resolvers with Same Method Names but Different Field Names Cause Schema Error
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

Multiple GraphQL Resolvers with Same Method Names but Different Field Names Cause Schema Error

### Minimum reproduction code

https://github.com/vietthang/nestjs-graphql-issue-reproduce

### Steps to reproduce

1. npm install
2. npm start
3. The application throws the following error:
```
Error: User.lastName defined in resolvers, but not in schema
```

### Expected behavior

The resolvers should be able to coexist since they resolve different fields (`firstName` and `lastName`) through the `name` property in their respective `@ResolveField` decorators.

### Package version

13.0.4

### Graphql version

`graphql`: 16.8.1
`apollo-server-express`: 3.13.0

### NestJS version

11.0.12

### Node.js version

20.18.1

### In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [ ] Linux

### Other

_No response_

### LINKED ISSUE #3778: Extensions are not applied to resolved fields with custom method names
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

Say that we defined a DTO like this:

```ts
@ObjectType()
export class User {
  @Field(() => ID)
  id: string;

  @Field()
  name: string;

  @Extensions({ isPublic: true })
  @Field(() => Status, { nullable: true, description: 'DTO Description' })
  status?: Status;
}
```

And we have the UserResolver defined like this:

```ts
@Resolver(() => User)
export class UserResolver {

  @ResolveField('status', undefined, {
    nullable: true,
    description: 'Resolve Field Description',
  })
  getStatus(@Parent() user: User): Status {
    return {
      id: 'status-id',
      code: 'ACTIVE',
    };
  }
}

```

When the GQL  Host Schema is generated, the `User.status` field does not have any extensions.

### Minimum reproduction code

https://github.com/thiagomini/input-field-no-extension-bug/tree/bug-with-resolve-field

### Steps to reproduce

1. `pnpm install`
2. `pnpm run start:extensions`

You should see the assertion failing:

```
AssertionError [ERR_ASSERTION]: Expected values to be loosely deep-equal:

[Object: null prototype] {
  complexity: undefined
}

should loosely deep-equal

{
  complexity: undefined,
  isPublic: true
}
    at bootstrap (/home/thiago/Development/roomvy/repros/input-field-no-extension-bug/src/print-extensions.ts:25:10) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: [Object: null prototype] { complexity: undefined },
  expected: { complexity: undefined, isPublic: true },
  operator: 'deepEqual',
  diff: 'simple'
}
```

### Expected behavior

The `User.status` field should have the `isPublic` extension key.

### Package version

13.2.3

### Graphql version

`graphql`: 16.12.0
`apollo/server`: 5.2.0

### NestJS version

11.1.9

### Node.js version

24.11.1

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [x] Linux

### Other

⚠️ Requires Attention - Possibly a breaking change

This bug only occurs when we use a **custom method name** for a resolve field. If we rename the method abovementioned to `status`, it works as expected.

However, it's worth noting that the underlying reason for this bug is how the `TypeMetadataStorage` class identifies a GQL field's metadata - when we rename a resolver's method name, it handles it as a **new** field . The fix I'll propose for this will probably fix #3524 too - but there's a catch here - this might impact how existing applications using the code-first approach build their schemas.

Currently, if we define an Object's field's metadata in **both** the DTO **and** in the @ResolveField() decorator, the `@ResolveField()` will take precedence (but that's a bug, as I mentioned, because of how the metadata storage thinks the method decorated with @ResolveField is actually a new field (when renamed). Let's recape the example:

```ts
@ObjectType()
export class User {
  @Field(() => ID)
  id!: string;

  @Field()
  name!: string;

  @Extensions({ isPublic: true })
  @Field(() => Status, { nullable: true, description: 'DTO Description' })
  status?: Status;
}
```

Notice the `status` field description: `DTO Description`.

Now, let's look at the @ResolveField definition:

```ts
@ResolveField('status', undefined, {
    nullable: true,
    description: 'Resolve Field Description',
  })
  getStatus(@Parent() user: User): Status {
    return {
      id: 'status-id',
      code: 'ACTIVE',
    };
  }
```

Right now, when our application schema is generated, the final description will be `Resolve Field Description`. But if we fix the existing bug, the field description will be updated to `DTO Description`.

I wanted to mention this because there might be several applications right now that will have their schema definitions changed if there are any discrepancies between the DTO and the @ResolveField, so, Kamil, you probably need to decide whether to apply this patch to a new major version or a hotfix.

thiagomini:
Closes #3778

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #3778

## What is the new behavior?

Extensions are correctly applied to resolved fields with renamed method names.

- [x] Yes (I think)
- [ ] No

## Other information

⚠️ Requires Attention - Possibly a breaking change

This bug only occurs when we use a **custom method name** for a resolve field. If we rename the method mentioned above to `status`, it works as expected.

However, it's worth noting that the underlying reason for this bug is how the `TypeMetadataStorage` class identifies a GQL field's metadata - when we rename a resolver's method name, it handles it as a **new** field . The fix I'll propose for this will probably fix #3524 too - but there's a catch here - this might impact how existing applications using the code-first approach build their schemas.

Currently, if we define an Object's field's metadata in **both** the DTO **and** in the @ResolveField() decorator, the `@ResolveField()` will take precedence (but that's a bug, as I mentioned, because of how the metadata storage thinks the method decorated with @ResolveField is actually a new field when renamed). Let's recap the example:

```ts
@ObjectType()
export class User {
  @Field(() => ID)
  id!: string;

  @Field()
  name!: string;

  @Extensions({ isPublic: true })
  @Field(() => Status, { nullable: true, description: 'DTO Description' })
  status?: Status;
}
```

Notice the `status` field description: `DTO Description`.

Now, let's look at the @ResolveField definition:

```ts
@ResolveField('status', undefined, {
    nullable: true,
    description: 'Resolve Field Description',
  })
  getStatus(@Parent() user: User): Status {
    return {
      id: 'status-id',
      code: 'ACTIVE',
    };
  }
```

Right now, when our application schema is generated, the final description will be `Resolve Field Description`. But if we fix the existing bug, the field description will be updated to `DTO Description`.

I wanted to mention this because there might be several applications right now that will have their schema definitions changed if there are any discrepancies between the DTO and the @ResolveField, so, Kamil, you probably need to decide whether to apply this patch to a new major version or a hotfix.

### DIFF
--- a/packages/apollo/tests/e2e/code-first-extensions.spec.ts
+++ b/packages/apollo/tests/e2e/code-first-extensions.spec.ts
@@ -1,0 +1,35 @@
+import { INestApplication } from '@nestjs/common';
+import { GraphQLSchemaHost } from '@nestjs/graphql';
+import { Test } from '@nestjs/testing';
+import * as assert from 'assert';
+import { GraphQLObjectType } from 'graphql';
+import { AppModule } from '../code-first-extensions/app.module';
+
+describe('Code-first extensions', () => {
+  let app: INestApplication;
+
+  beforeAll(async () => {
+    const moduleRef = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+
+    app = moduleRef.createNestApplication();
+    await app.init();
+  });
+
+  afterAll(async () => {
+    await app.close();
+  });
+
+  it('Adds extensions to resolved field', async () => {
+    const { schema } = app.get(GraphQLSchemaHost);
+    const userObject = schema.getType('User') as GraphQLObjectType;
+    assert(userObject, 'User type not found in schema');
+    const statusField = userObject.getFields().status;
+    assert(statusField, 'status field not found in User type');
+
+    const extensions = statusField.extensions;
+
+    expect(extensions.isPublic).toBe(true);
+  });
+});
--- a/packages/graphql/lib/schema-builder/storages/type-metadata.storage.ts
+++ b/packages/graphql/lib/schema-builder/storages/type-metadata.storage.ts
@@ -421,7 +421,7 @@ export class TypeMetadataStorageHost {
     }
     let objectOrInterfaceTypeField =
       objectOrInterfaceTypeMetadata.properties.find(
-        (fieldDef) => fieldDef.name === item.methodName,
+        (fieldDef) => fieldDef.schemaName === item.schemaName,
       );
     for (
       let _objectTypeRef = objectTypeRef;
@@ -430,7 +430,7 @@ export class TypeMetadataStorageHost {
     ) {
       const possibleTypeMetadata = getTypeMetadata(_objectTypeRef);
       objectOrInterfaceTypeField = possibleTypeMetadata?.properties.find(
-        (fieldDef) => fieldDef.name === item.methodName,
+        (fieldDef) => fieldDef.schemaName === item.schemaName,
       );
       if (objectOrInterfaceTypeField) {
         objectTypeRef = _objectTypeRef;
--- a/packages/apollo/tests/code-first-extensions/app.module.ts
+++ b/packages/apollo/tests/code-first-extensions/app.module.ts
@@ -1,0 +1,19 @@
+import { GraphQLModule } from '@nestjs/graphql';
+import { ApolloDriver, ApolloDriverConfig } from '../../lib';
+import { UserModule } from './user/user.module';
+import { Module } from '@nestjs/common';
+
+/**
+ * Main application module for the code-first extensions example.
+ */
+@Module({
+  imports: [
+    GraphQLModule.forRoot<ApolloDriverConfig>({
+      driver: ApolloDriver,
+      autoSchemaFile: true,
+      playground: false,
+    }),
+    UserModule,
+  ],
+})
+export class AppModule {}
--- a/packages/apollo/tests/code-first-extensions/user/create-user.input.ts
+++ b/packages/apollo/tests/code-first-extensions/user/create-user.input.ts
@@ -1,0 +1,9 @@
+import { Extensions, Field, InputType } from '@nestjs/graphql';
+
+@InputType()
+@Extensions({ exampleExtension: 'exampleValue' })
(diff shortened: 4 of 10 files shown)


packages/apollo/tests/e2e/code-first-extensions.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    app = moduleRef.createNestApplication();
+    await app.init();
+  });
+
+  afterAll(async () => {
+    await app.close();
+  });
+
+  it('Adds extensions to resolved field', async () => {
```
thiagomini: Note: I've asserted that this test fails without the changed I made to the type-metadata.storage

packages/graphql/lib/schema-builder/storages/type-metadata.storage.ts
```diff
@@ -421,7 +421,7 @@ export class TypeMetadataStorageHost {
     }
     let objectOrInterfaceTypeField =
       objectOrInterfaceTypeMetadata.properties.find(
-        (fieldDef) => fieldDef.name === item.methodName,
+        (fieldDef) => fieldDef.schemaName === item.schemaName,
```
thiagomini: I don't know why this didn't compare `schemaName` to find existing fields' metadata. When comparing to `methodName` it wrongly assumes the methodName will always reflect the schemaName, which is **not** the case.

PR #3800, opened 2026-01-18
fix(graphql): resolve scalar type mapping failures with Jest globalSetup

### LINKED ISSUE #3356: GraphQL Nested types + JEST bug
### Is there an existing issue for this?

### Current behavior

Application works in development environment without problems, yet in `e2e` tests particularly on the second `app.init()` i.e. after `globalSetup` runs it throws
`Cannot determine a GraphQL output type for the "text". Make sure your class is decorated with an appropriate decorator.`.

This happens only when I have a nested model like [this](https://github.com/G0maa/i18n-graphql/blob/main/src/dto/responses.dto.ts) one.

### Minimum reproduction code

https://github.com/G0maa/i18n-graphql

### Steps to reproduce

1. `npm install`
2. Verify that development works `npm run start:dev`
3. `npm run test:e2e`
- or just visit [this](https://gitpod.io/?autostart=true#https://github.com/G0maa/i18n-graphql) GitPod link and it will do everything for you :smile: 

### Expected behavior

Tests should work without throwing the error specified above.

### Package version

12.2.1

### Graphql version

    "@apollo/server": "^4.11.0",
    "@nestjs/apollo": "^12.2.1",
    "graphql": "^16.9.0",

### NestJS version

10.4.6

### Node.js version

v20.17.0

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

### Other

I've spent many ~~hours~~ days searching for a solution, I've also asked on [Discord](https://discord.com/channels/520622812742811698/1296441519221506111/1296441519221506111), I am yet to find a solution to this bug. There's a similar [issue](https://github.com/nestjs/graphql/issues/2393), which its owner also posted on Discord and as it appears he didn't find a solution either.

joe-re:
Jest globalSetup runs in a separate VM context. Built-in constructor references (String, Number, Boolean) differ between contexts, which breaks Map-based lookups using strict equality.

Use name-based lookup in TypeMapperService.mapToScalarType() and prevent duplicate ObjectType registration by name in TargetMetadataCollection.

Closes #3356

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When using Jest's globalSetup to pre-initialize a NestJS GraphQL application, ObjectTypes with scalar fields (String, Number, Boolean, Date) fail with the error:

```
Cannot determine a GraphQL output type for the 'fieldName'. Make sure your class is decorated with an appropriate decorator.
```

This occurs because Jest's globalSetup runs in a separate VM context from test files(https://jestjs.io/docs/configuration#globalsetup-string, https://github.com/jestjs/jest/issues/6007).
Built-in constructor references captured by @Field() decorators in globalSetup differ from those in the test context, causing TypeMapperService.mapToScalarType() Map-based lookups with strict equality (===) to fail.

Issue Number: #3356

## What is the new behavior?

- Scalar types are correctly mapped regardless of which VM context the type reference originated from
- globalSetup can be used safely with NestJS GraphQL applications
- Duplicate ObjectType registrations from different contexts are handled by replacing existing entries with the same name

Changes:
- TypeMapperService.mapToScalarType(): Use name-based lookup (typeRef.name) instead of Map-based reference equality
- TargetMetadataCollection.objectType setter: Prevent duplicate registrations by checking type name before adding

- [x] Yes
- [ ] No

The change from reference equality to name-based lookup introduces a minor behavioral difference. If a user defines a custom class with the same name as a built-in type (e.g., class String {}), it will now be incorrectly mapped to the corresponding GraphQL scalar.

This is considered low risk because:
- Naming custom classes the same as built-in types is generally bad practice
- TypeScript/ESLint typically warns against shadowing built-in types
- Such usage is extremely rare in practice

## Other information

Calling TypeMetadataStorage.clear() at the end of globalSetup also resolves the issue without library code changes, but requires users to know about and implement this workaround. 
The name-based approach provides a seamless experience.
I'm not sure which is the best, so I'll leave the judgment to you all.

### DIFF
--- a/packages/apollo/package.json
+++ b/packages/apollo/package.json
@@ -15,7 +15,7 @@
     "url": "git+https://github.com/nestjs/graphql.git"
   },
   "scripts": {
-    "test:e2e": "jest --config ./tests/jest-e2e.ts --runInBand",
+    "test:e2e": "jest --config ./tests/jest-e2e.ts --runInBand && jest --config ./tests/global-setup-issue/jest-global-setup.ts --runInBand",
     "test:e2e:dev": "jest --config ./tests/jest-e2e.ts --runInBand --watch"
   },
   "bugs": {
--- a/packages/apollo/tests/global-setup-issue/app.module.ts
+++ b/packages/apollo/tests/global-setup-issue/app.module.ts
@@ -1,0 +1,15 @@
+import { Module } from '@nestjs/common';
+import { GraphQLModule } from '@nestjs/graphql';
+import { ApolloDriver, ApolloDriverConfig } from '../../lib';
+import { PostsResolver } from './posts.resolver';
+
+@Module({
+  imports: [
+    GraphQLModule.forRoot<ApolloDriverConfig>({
+      driver: ApolloDriver,
+      autoSchemaFile: true,
+    }),
+  ],
+  providers: [PostsResolver],
+})
+export class GlobalSetupIssueModule {}
--- a/packages/apollo/tests/global-setup-issue/global-setup-issue.e2e-spec.ts
+++ b/packages/apollo/tests/global-setup-issue/global-setup-issue.e2e-spec.ts
@@ -1,0 +1,90 @@
+import { INestApplication } from '@nestjs/common';
+import { GraphQLModule } from '@nestjs/graphql';
+import { Test, TestingModule } from '@nestjs/testing';
+import { ApolloServer } from '@apollo/server';
+import { gql } from 'graphql-tag';
+import { ApolloDriver } from '../../lib';
+import { GlobalSetupIssueModule } from './app.module';
+import { expectSingleResult } from '../utils/assertion-utils';
+
+/**
+ * This test reproduces the issue described in:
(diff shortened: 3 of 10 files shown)


joe-re:
After creating this PR and investigating the CI failures, I discovered that this issue is more complex than initially anticipated.

The name-based lookup approach in TypeMapperService fixes the scalar type mapping issue, but it doesn't address the ObjectType duplication problem. When the same ObjectType is registered from both the globalSetup VM context and the test file VM context, GraphQL throws:

Schema must contain uniquely named types but contains multiple types named "TypeName"

I attempted to fix this by adding duplicate detection in TargetMetadataCollection, but this broke the Mercurius federation tests due to a fundamental data structure mismatch:
- metadataByTargetCollection uses target references as keys
- all.objectType requires unique names

When two different target references (from different VM contexts) represent the same logical type, fixing one breaks the other.

For now, I believe the best approach is to document the workaround using TypeMetadataStorage.clear() on globalSetup.

I will make a PR for the documentation.
Happy to keep discussing better approaches.

Issue #3025, opened 2023-10-13
Enums with one key are treated as String literals


IodizedGabe:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When defining a field for an object, using an enum with exactly 1 key will result in that field being seen as a String, rather than an enum.

The type information for this field sets its `flags` as 1152, or 1024 | 128, or [EnumLiteral](https://github.com/microsoft/TypeScript/blob/cf8763b6c3193deefd095c86ce79921f23de307b/src/compiler/types.ts#L6053 ) | [StringLiteral](https://github.com/microsoft/TypeScript/blob/cf8763b6c3193deefd095c86ce79921f23de307b/src/compiler/types.ts#L6050).

When it's then parsed, it will check if it's a string [first](https://github.com/nestjs/graphql/blob/master/packages/graphql/lib/plugin/utils/plugin-utils.ts#L49-L50) before checking if it's an enum.

That said, even if this were then checked as an enum, it will probably fail [here](https://github.com/nestjs/graphql/blame/master/packages/graphql/lib/plugin/utils/ast-utils.ts#L76-L77).

### Minimum reproduction code

https://github.com/IodizedGabe/nestjs-graphql-enum-as-string

### Steps to reproduce

1. `yarn install`
2. `yarn build`
3. Look at the `_GRAPHQL_METADATA_FACTORY` for `ClassA` in `./dist/graphql.entity.js`

For me, I see:
```
let ClassA = class ClassA {
    static _GRAPHQL_METADATA_FACTORY() {
        return { multipleEnum: { type: () => require("./graphql.entity").MultipleEnum }, singleEnum: { type: () => String } };
    }
};
```

### Expected behavior

An enum with one key should be treated like any other enum and given a proper type / reference in the schema.

### Package version

12.0.9

### Graphql version

`graphql`: 16.8.1

### NestJS version

10.2.7

### Node.js version

18.15.0

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [ ] Linux

### Other

`typescript`: 5.1.6

kamilmysliwiec:
Would you like to create a PR for this issue?

IodizedGabe:
I could, but I'm not exactly sure of the best way to work around this. The simplest change would be to add another flag check [here](https://github.com/nestjs/graphql/blob/master/packages/graphql/lib/plugin/utils/ast-utils.ts#L55), but I'm not sure if that's the best way to go about it.

[...]

PR #3934, opened 2026-04-18
fix(@nestjs/graphql): treat single-key string enums as enums in plugin type detection

maruthang:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Bugfix.

Issue Number: #3025

When a property is typed as a single-member string enum, the plugin's CLI metadata generator emits the property as `String` instead of the enum. The `isStringLiteral` helper in `packages/graphql/lib/plugin/utils/ast-utils.ts` matches any type whose flags include `StringLiteral`, but a one-key string enum carries both the `StringLiteral` and `EnumLiteral` flags, so the enum branch is never reached and `type: () => Enum` metadata is lost.

## What is the new behavior?

`isStringLiteral` now also excludes the `EnumLiteral` flag, so single-key enum members fall through to the existing enum branch and are emitted with the correct `type: () => Enum` metadata. Multi-key string enums were already handled because they surface as unions, which the existing check excluded. A regression test was added under `packages/graphql/tests/plugin/model-class-visitor.spec.ts` that asserts a single-key string enum property is detected as an enum rather than a string literal.

- [ ] Yes
- [x] No

## Other information

Verified locally with `yarn test:e2e:graphql` (87 passed, 1 skipped). Two snapshot tests fail only on Windows due to a pre-existing CRLF/LF fixture issue that is unrelated to this change.

### DIFF
--- a/packages/graphql/lib/plugin/utils/ast-utils.ts
+++ b/packages/graphql/lib/plugin/utils/ast-utils.ts
@@ -53,7 +53,11 @@ export function isString(type: Type) {
 }
 
 export function isStringLiteral(type: Type) {
-  return hasFlag(type, TypeFlags.StringLiteral) && !type.isUnion();
+  return (
+    hasFlag(type, TypeFlags.StringLiteral) &&
+    !hasFlag(type, TypeFlags.EnumLiteral) &&
+    !type.isUnion()
+  );
 }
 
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

Issue #3180, opened 2024-03-10
Unable to Specify forRoutes Argument When Assigning Middleware to '/graphql' Endpoint


stringthread:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Using `app.setGlobalPrefix` and `GraphQLModule.forRoot`, I cannot specify `/graphql` endpoint in the argument of `consumer.apply(middleware).forRoutes`.

```typescript
// main.ts
import { NestFactory } from '@nestjs/core';
import { AppModule } from './app.module';

async function bootstrap() {
  const app = await NestFactory.create(AppModule);
  app.setGlobalPrefix('prefix');
  await app.listen(3000);
}
bootstrap();
```
```typescript
// app.module.ts
import { Module, NestModule, MiddlewareConsumer } from '@nestjs/common';
import { GraphQLModule } from '@nestjs/graphql';
import { ApolloDriver } from '@nestjs/apollo';
import { AppResolver } from './app.resolver'; // Simple sample resolver
import * as path from 'path';

@Module({
  imports: [
    GraphQLModule.forRoot({
      driver: ApolloDriver,
      autoSchemaFile: path.join(process.cwd(), 'src/schema.gql'),
    }),
  ],
  providers: [AppResolver],
})
export class AppModule implements NestModule {
  configure(consumer: MiddlewareConsumer) {
    consumer
      .apply((req, res, next) => {
        console.log('apply');
        next();
      })
      .forRoutes('graphql'); // This don't work
  }
}
```

### Minimum reproduction code

https://codesandbox.io/p/devbox/nestjs-graphql-middleware-issue-8d6qfr

### Steps to reproduce

1. `npm start:dev`
2. access `/graphql` endpoint
3. See terminal outputs

### Expected behavior

The middleware should be called when I access `/graphql` endpoint.
In my minimum example, `apply` should be shown in terminal outputs.

### Package version

12.1.1

### Graphql version

`graphql`: 16.8.1
`@apollo/server`: 4.10.1
`@nestjs/apollo`: 12.1.0

### NestJS version

10.3.3

### Node.js version

20.11.0

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

### Other

I have found a workaround to avoid this bug:
`consumer.apply(middleware).forRoutes` worked properly when I specified `exclude` option in `app.setGlobalPrefix` and add "graphql".

```typescript
// main.ts
async function bootstrap() {
[...]

PR #3935, opened 2026-04-19
fix(@nestjs/graphql): exclude graphql path from global prefix by default

maruthang:
## PR Checklist

- [x] The commit message follows our guidelines: https://github.com/nestjs/graphql/blob/master/CONTRIBUTING.md
- [x] Docs have been added / updated (for bug fixes / features) — N/A (behavior fix; no public API change)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Note: only `Bugfix` is ticked above.

- [x] Bugfix

Issue Number: #3180

Middleware registered with `consumer.apply(SomeMiddleware).forRoutes('graphql')` does not fire for GraphQL requests when `app.setGlobalPrefix(...)` is set without `useGlobalPrefix: true` on the GraphQL module. NestJS auto-prepends the global prefix to user middleware paths, but the GraphQL endpoint stays at `/graphql`, so the two paths diverge (middleware ends up bound at `/prefix/graphql` while the endpoint serves `/graphql`).

## What is the new behavior?

`GraphQLModule` now auto-excludes its endpoint path from the application's global prefix on construction (only when `useGlobalPrefix` is not enabled). User middleware registered via `forRoutes('graphql')` now resolves to `/graphql` and matches the endpoint.

Implementation lives in `packages/graphql/lib/graphql.module.ts` (`excludeFromGlobalPrefix`); regression tests added under `packages/graphql/tests/global-prefix-middleware.spec.ts`. The DI snapshot in `packages/apollo/tests/e2e/__snapshots__/serialized-graph.spec.ts.snap` was regenerated to reflect the new `ApplicationConfig` constructor edge.

- [ ] Yes
- [x] No

## Other information

This is a default behavior change: prior to this PR, projects with `app.setGlobalPrefix('prefix')` (no `useGlobalPrefix` on the GraphQL module) had their middleware bound at `/prefix/graphql`, which never matched the GraphQL endpoint at `/graphql`. After this PR, the middleware binds at `/graphql` and fires correctly. Projects relying on the previous (broken) behavior should opt back in by setting `useGlobalPrefix: true` on `GraphQLModule.forRoot(...)`. Verified with `yarn test:e2e:graphql` and `yarn test:e2e:apollo` — all pre-existing global-prefix specs continue to pass.

### DIFF
--- a/packages/apollo/tests/e2e/__snapshots__/serialized-graph.spec.ts.snap
+++ b/packages/apollo/tests/e2e/__snapshots__/serialized-graph.spec.ts.snap
@@ -2547,6 +2547,23 @@ exports[`Serialized graph > should generate a post-initialization graph and matc
       },
       "id": "-1187631397"
     },
+    "-2131886386": {
+      "source": "-1497318880",
+      "target": "2022351939",
+      "metadata": {
+        "type": "class-to-class",
+        "sourceModuleName": "GraphQLModule",
+        "sourceClassName": "GraphQLModule",
+        "targetClassName": "ApplicationConfig",
+        "sourceClassToken": "GraphQLModule",
+        "targetClassToken": "ApplicationConfig",
+        "targetModuleName": "GraphQLModule",
+        "keyOrIndex": 5,
+        "injectionType": "constructor",
+        "internal": true
+      },
+      "id": "-2131886386"
+    },
     "-1983130485": {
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
Feels a bit implicit?

Issue #2960, opened 2023-08-10
`@Directive` on abstract `ObjectType` is not applied to inherited classes


NilsMoller:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Having a base `@ObjectType({ isAbstract: true })` class decorated with `@Directive('@key(fields: "id")')` does not apply that directive to inherited classes.

### Minimum reproduction code

N/A

### Steps to reproduce

1. Copy sample 31 for code-first federation.
2. Add a `model.base.ts` to `users-application/src/users/models`:
```ts
@ObjectType({
  isAbstract: true,
})
@Directive('@key(fields: "id")')
export class ModelBase {
  @Field((type) => ID)
  id: number;
}
```
3. Update the `user.model.ts`:
```ts
@ObjectType()
export class User extends ModelBase {
  @Field()
  name: string;
}
```
4. Start the projects (gateway last)
5. See the error in the gateway console

### Expected behavior

The directive is applied to all inheriting classes.

### Package version

12.0.8

### Graphql version

```
"@apollo/gateway": "2.2.3",
"@apollo/server": "4.7.5",
"@nestjs/apollo": "12.0.3",
"@nestjs/graphql": "12.0.3",
"graphql": "16.6.0",
"graphql-tools": "9.0.0",
```

### NestJS version

^10.0.3

### Node.js version

18.17.0

### In which operating systems have you tested?

- [ ] macOS
- [X] Windows
- [ ] Linux

### Other

- Tried with both federation 1 and 2.
- I presume this is an issue not specifically related to federation, but I do not know enough about directives to test other configurations for now.
- I have tested this with the (currently) latest versions in my own project with the same results:
```
"@apollo/server": "^4.9.1",
"@nestjs/apollo": "^12.0.7",
"@nestjs/graphql": "^12.0.8",
"graphql": "^16.7.1",
```

kamilmysliwiec:
Would you like to create a PR for this issue?

PR #3938, opened 2026-04-19
fix(@nestjs/graphql): inherit class directives from abstract parents

maruthang:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Only `Bugfix` applies: [x] Bugfix.

Issue Number: #2960

When a `@Directive(...)` decorator is placed on an abstract `@ObjectType()` parent class (e.g. `@Directive('@key(fields: "id")')`), classes that extend that parent do not inherit the directive metadata. Field metadata inherits correctly, but class-level directives silently disappear on every derived concrete type. The bug surfaces in federation setups where shared keys/directives live on a base class.

## What is the new behavior?

`TypeMetadataStorageHost.compileClassMetadata()` now sources class-level directive metadata via a new private helper, `getInheritedClassDirectives(target)`, which walks the prototype chain via `Object.getPrototypeOf` and collects `classDirectives` from every ancestor, de-duplicating by SDL string. Child-first ordering means own directives win the de-dup pass, preserving prior single-class behavior.

A regression spec at `packages/graphql/tests/schema-builder/factories/directive-inheritance.spec.ts` declares an abstract `@ObjectType({ isAbstract: true })` parent decorated with `@Directive('@key(fields: "id")')`, a derived concrete `@ObjectType()`, builds the schema via `GraphQLSchemaFactory`, and asserts the resulting type's `astNode.directives` includes `key`. (graphql-js `printSchema` does not emit type-level directives in SDL — that's why the assertion targets `astNode.directives`, which is what federation tooling reads.)

- [ ] Yes
- [x] No

The change strictly adds inherited directives to the compiled metadata; classes that previously had no inherited directives still have none if no ancestor declared any. Existing classes that declared their own directives continue to see only those directives (de-dup keeps the child copy first).

## Other information

Verified with `yarn test:e2e:graphql` (87 passed plus the unrelated pre-existing Windows CRLF snapshot failures in `tests/plugin/model-class-visitor.spec.ts` `es5-eager-imports` and `tests/plugin/readonly-visitor.spec.ts` `should generate a serialized metadata`, both reproducible on clean master). `npx oxlint` and `npx prettier` clean on touched files only.

### DIFF
--- a/packages/graphql/lib/schema-builder/storages/type-metadata.storage.ts
+++ b/packages/graphql/lib/schema-builder/storages/type-metadata.storage.ts
@@ -282,9 +282,7 @@ export class TypeMetadataStorageHost {
         item.properties = this.getClassFieldsByPredicate(item);
       }
       if (!item.directives) {
-        item.directives = this.metadataByTargetCollection
-          .get(item.target)
-          .classDirectives.getAll();
+        item.directives = this.getInheritedClassDirectives(item.target);
       }
       if (!item.extensions) {
         item.extensions = this.metadataByTargetCollection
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
image

maruthang:
Thanks for the screenshot @kamilmysliwiec — reproduced locally. The issue was that my first implementation built a fresh `directives` array inside `getInheritedClassDirectives()`, which broke an implicit contract: `compile()` reverses the per-target `classDirectives` array in place, and `item.directives` needs to share that reference so repeated `compile()` invocations stay consistent across the run. Copying into a new array froze the order at the first invocation, which is what you saw in the `User` federation type.

Pushed 559369f — when a target has no parent directives to inherit, we return the original reference (identical to the pre-PR behavior). We only allocate a new array when we actually have to merge in directives from an abstract parent.

Local verification:
- `packages/mercurius` — 21 files / 39 tests pass (incl. `code-first-federation.spec.ts`)
- `packages/apollo` — 39 files / 129 tests pass
- `packages/graphql` — 15 files pass + the new `directive-inheritance.spec.ts` (the 2 plugin-related failures also fail on `master` without my change)

CI should be green on the next run.

Issue #2587, opened 2023-01-10
ResolveField ignores field middleware


equal-matt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When resolving a field with `@ResolveField`, any middleware defined on the  corresponding `ObjectType`'s `@Field` definition is ignored.

This is particularly annoying when a subset of fields for an object are resolved. Without also finding the resolver for the field that you're looking at, there is no guarantee that your middleware will execute as specified. To handle middleware-like transformations consistently, you are forced to handle all middleware-like transformations in field resolvers by hand.

### Minimum reproduction code

https://stackblitz.com/edit/nestjs-typescript-starter-rkgnvv?file=src/example/example.resolver.ts&view=editor

### Steps to reproduce

Use the GraphQL playground in the preview section to query for

```gql
query {
  allExamples {
    fieldMiddlewareField
    resolveMiddlewareField
    bothMiddlewareField
  }
}
```

### Expected behavior

Middleware defined in an `ObjectType`'s `@Field` decorator should apply to the returned values from the `@ResolveField` handler.

### Package version

10.1.7

### Graphql version

```
"@nestjs/apollo": "^10.1.7",
"@nestjs/graphql": "^10.1.7",
"@nestjs/platform-express": "^9.0.0",
"apollo-server-express": "^3.11.1",
"graphql": "^16.6.0",
```

### NestJS version

9.2.1

### Node.js version

16.14.2

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

choco14t:
@equal-matt 
I tried to fix `example.resolver.ts`, field middleware works correctly.
I don't know details, it seems that returnType in `@ResolveField` needs to define type.

```typescript
import { Query, ResolveField, Resolver } from '@nestjs/graphql';
import { Example } from './example.model';

@Resolver((_) => Example)
export class ExampleResolver {
  @Query(() => [Example])
  allExamples() {
    return [{ fieldMiddlewareField: '  will be trimmed  ' }];
  }

  @ResolveField(() => String, {
    middleware: [async (_, next) => (await next()).trim()],
  })
  resolveMiddlewareField() {
    return "  won't be trimmed  ";
  }

  @ResolveField(() => String, {
    middleware: [async (_, next) => (await next()).trim()],
  })
  bothMiddlewareField() {
    return "  still won't be trimmed  ";
  }
}
```

equal-matt:
[...]

PR #3939, opened 2026-04-19
fix(@nestjs/graphql): preserve ResolveField options for all overloads

maruthang:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

(Marking only Bugfix:)

- [x] Bugfix

Issue Number: #2587

Field middleware configured via `@ResolveField({ middleware: [...] })` or `@ResolveField('aliased', { middleware: [...] })` is silently dropped. The same middleware fires correctly for plain `@Field({ middleware: [...] })` and for `@ResolveField(() => SomeType, { middleware: [...] })`. The reporter and a co-reporter on the original issue identified that adding an explicit `returnType` argument to `@ResolveField` makes the middleware work — that observation is the fingerprint of the underlying decorator-argument-destructuring bug, not a middleware-wiring problem.

## What is the new behavior?

`packages/graphql/lib/decorators/resolve-field.decorator.ts` now branches on the actual shape of the arguments rather than only on `isFunction(firstArg)`. New routing:

(a) first arg function → typeFunc-first overload (unchanged);
(b) first arg string + second arg function → propertyName + typeFunc + options (unchanged);
(c) first arg string + second arg object → propertyName + no-typeFunc + options;
(d) first arg object → no-name + no-typeFunc + options.

Adds an `isString` import alongside the existing `isFunction`/`isObject` from `@nestjs/common/utils/shared.utils`.

A regression spec at `packages/graphql/tests/schema-builder/factories/resolve-field-middleware.spec.ts` reads `FIELD_RESOLVER_MIDDLEWARE_METADATA`, `RESOLVER_NAME_METADATA`, and `RESOLVER_PROPERTY_METADATA` straight off the decorated method for three cases (typeFunc + options, options only, propertyName + options) — reading metadata directly avoids requiring an apollo/mercurius runtime to assert the wiring.

- [ ] Yes
- [x] No

The change only adds correct handling for the previously-broken overloads. Calls that already worked (typeFunc-first or propertyName + typeFunc + options) continue to produce identical metadata.

## Other information

Verified with `yarn test:e2e:graphql` (89 passed plus the unrelated pre-existing Windows CRLF snapshot failures in `tests/plugin/model-class-visitor.spec.ts` `es5-eager-imports` and `tests/plugin/readonly-visitor.spec.ts`, both reproducible on clean master). `npx oxlint` clean on touched files. Prettier formatting applied with the project's bundled v3 binary (the repo's `npx prettier` shim resolves to v2 and would reformat unrelated lines).

### DIFF
--- a/packages/graphql/lib/decorators/resolve-field.decorator.ts
+++ b/packages/graphql/lib/decorators/resolve-field.decorator.ts
@@ -1,5 +1,9 @@
 import { SetMetadata, Type } from '@nestjs/common';
-import { isFunction, isObject } from '@nestjs/common/utils/shared.utils';
+import {
+  isFunction,
+  isObject,
+  isString,
+} from '@nestjs/common/utils/shared.utils';
 import {
   FIELD_RESOLVER_MIDDLEWARE_METADATA,
   RESOLVER_NAME_METADATA,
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

Issue #2940, opened 2023-07-27
GraphQL error not formatted properly when it returns a 4xx or 5xx http status


remonke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

In Apollo, there's a way of providing a custom http status via `extensions` on the `GraphQLError` / `ApolloError`, as is the case with the default validation error, which returns a status of 400. Unfortunately, instead of the response being formatted the usual way - with `data` and `errors`, everything gets wrapped inside of the `error` object. Is there any way to disable this behavior?

![Screenshot_20230727_050425](https://github.com/nestjs/graphql/assets/102010617/420a7679-2808-4f21-a686-ba891d357c68)
![Screenshot_20230727_050459](https://github.com/nestjs/graphql/assets/102010617/bb701a20-9582-4e11-9559-d77bbb991241)

### Minimum reproduction code

https://github.com/remonke/nest-graphql-error-bug

### Steps to reproduce

Install the dependencies `pnpm install`
Start the server
Open the playground
Test the two available queries `throwError` and `throwErrorWithHttpStatus`

### Expected behavior

I expect a GraphQL errror with a custom http status to get formatted like all the others.

### Package version

12.0.8

### Graphql version

`graphql`: `16.7.1`
`@apollo/server`: `4.8.1`
`@nestjs/apollo`: `^12.0.7`

### NestJS version

10.0.0

### Node.js version

20.4.0

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

kamilmysliwiec:
Would you like to create a PR for this issue?

remonke:
I would try, except I have no idea what causes this behavior - I tried this with Mercurius driver and it works just fine

jan-hecker:
You can probably disable this behavior if you create a custom error formatter. You can provide the error formatter in the GraphQL options with the `formatError` argument

tanmen:
> You can probably disable this behavior if you create a custom error formatter. You can provide the error formatter in the GraphQL options with the formatError argument

I have tried this and it did not resolve the problem.

PR #3940, opened 2026-04-19
fix(apollo): preserve HTTP 200 for execution-level GraphQL errors

maruthang:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

(Selected: **Bugfix**.)

Issue Number: #2940

When a resolver throws a raw `GraphQLError` whose `extensions.http.status` is set (or a NestJS HTTP exception that maps onto Apollo's HTTP-status semantics), the GraphQL endpoint returns the resolver-derived HTTP 4xx/5xx status. Some GraphQL clients and UIs (Apollo Studio, Sandbox) then refuse to parse the response as `{data, errors}` and wrap it as a transport-level `{ error: ... }` envelope, which is what the issue's reporter sees. The bug is Apollo-specific because Mercurius does not honour `extensions.http.status`.

## What is the new behavior?

Adds a tiny Apollo Server plugin (`createPreserveHttpStatusPlugin`) registered in `ApolloBaseDriver.mergeDefaultOptions`. The plugin runs in `willSendResponse` after Apollo has finished merging error-derived HTTP status onto the response and resets the status to **200** when the request reached execution (detected by the presence of a `data` key on `singleResult`). Request-level failures (parse, validate) are untouched because their body has no `data` key. The plugin is gated on the existing `autoTransformHttpErrors` flag — users who want Apollo's default behaviour can opt out with `autoTransformHttpErrors: false`. Regression spec at `packages/apollo/tests/e2e/issue-2940-http-error-status.spec.ts` covers four cases: NestJS exception → 200, raw GraphQLError with `http.status: 400` → 200, parse failure → 400 preserved, opt-out → 400 preserved.

- [ ] Yes
- [x] No

Users currently throwing raw `GraphQLError` with `extensions.http.status` set deliberately and *expecting* the HTTP status to leak into the response will now see HTTP 200 unless they set `autoTransformHttpErrors: false`. The flag already exists in the public driver config and is exactly the knob for this behaviour.

## Other information

The plugin reuses the existing public flag `autoTransformHttpErrors`. Users who explicitly relied on `extensions.http.status` mapping to a non-2xx response can opt back in by setting `autoTransformHttpErrors: false`. Verified with the full apollo suite (`yarn test:e2e:apollo`: 133 passed / 3 skipped) including existing `pipes.spec.ts` and `guards-filters.spec.ts` error-handling tests, plus `yarn test:e2e:graphql` (only the unrelated pre-existing Windows CRLF snapshot failures in `tests/plugin/model-class-visitor.spec.ts` `es5-eager-imports` and `tests/plugin/readonly-visitor.spec.ts` appear; both reproducible on clean master). Prettier (v3 binary) and oxlint clean on touched files.

### DIFF
--- a/packages/apollo/lib/drivers/apollo-base.driver.ts
+++ b/packages/apollo/lib/drivers/apollo-base.driver.ts
@@ -103,6 +103,13 @@ export abstract class ApolloBaseDriver<
 
     this.wrapContextResolver(options);
     this.wrapFormatErrorFn(options);
+
+    if (options.autoTransformHttpErrors !== false) {
+      (options as ApolloDriverConfig).plugins = [
+        ...((options as ApolloDriverConfig).plugins || []),
+        this.createPreserveHttpStatusPlugin(),
+      ];
+    }
     return options;
   }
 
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

PR #3948, opened 2026-04-22
feat(@nestjs/graphql): support directives on enums and unions

### LINKED ISSUE #2920: Add capability to add directive to GraphQL enum and union (Code First approach)

Currently, even though the GraphQL spec allow directive to be added to enum and union. There is no possible solutions at the moment to add them to Enum and Union with the code first approach. 

For enum, I think we can add a `directive` options to`registerEnumType`
`registerEnumType(COLOR, {
  valuesMap: {
    RED: {
      description: 'Risk record',
      directive: '@directive'
    },
    BLUE: {
      description: 'Business service record in OpRes',
      directive: '@directive'
    },
    directive:  '@directive'
});`
The same thing can be done for `createUnionType`:
createUnionType({ name: 'UnionName', types: ..., directive: '@directive' })

_No response_

We need the ability to apply a directive to types and fields in our GraphQL schema and then at build time generate different schema for different environments (parner/third-party and internal use). This is similar to [Apollo Contract](https://www.apollographql.com/docs/graphos/delivery/contracts/) . Even if we use Apollo Contract, we need a way to apply the `@tag` directive to enum, enum values and union to hide it.

yogeshwaran-c:
## Summary
- Add a `directives` option to `registerEnumType`, `createUnionType`, and enum `valuesMap` entries to attach custom directives (e.g. `@tag`, `@inaccessible`) via SDL strings.
- Factories now build the corresponding AST nodes so the directives round-trip through `printSchemaWithDirectives` and tools such as Apollo Contracts.
- Closes #2920.

## Motivation
The GraphQL spec allows directives on enum types, enum values, and union types. Today there is no supported way to attach them with the code-first API, which blocks use cases like Apollo Contracts (`@tag` on enums/unions to hide them from partner schemas).

## Example
```ts
enum MovieGenre { ACTION = 'ACTION', COMEDY = 'COMEDY' }
registerEnumType(MovieGenre, {
  name: 'MovieGenre',
  directives: ['@tag(name: "public")'],
  valuesMap: { ACTION: { directives: ['@tag(name: "featured")'] } },
});

const Media = createUnionType({
  name: 'Media',
  types: () => [Book, Movie] as const,
  directives: ['@tag(name: "public")'],
});
```

## Test plan
- [x] New spec `enum-union-directives.spec.ts` verifies directives on the enum type, union type, and a single enum value.
- [x] Existing `tests/schema-builder/**` suites still pass.

### DIFF
--- a/packages/apollo/tests/e2e/__snapshots__/serialized-graph.spec.ts.snap
+++ b/packages/apollo/tests/e2e/__snapshots__/serialized-graph.spec.ts.snap
@@ -1642,6 +1642,22 @@ exports[`Serialized graph > should generate a post-initialization graph and matc
       },
       "id": "1118498651"
     },
+    "1159932337": {
+      "source": "576628947",
+      "target": "-1155703408",
+      "metadata": {
+        "type": "class-to-class",
+        "sourceModuleName": "GraphQLSchemaBuilderModule",
+        "sourceClassName": "EnumDefinitionFactory",
+        "targetClassName": "AstDefinitionNodeFactory",
+        "sourceClassToken": "EnumDefinitionFactory",
+        "targetClassToken": "AstDefinitionNodeFactory",
+        "targetModuleName": "GraphQLSchemaBuilderModule",
+        "keyOrIndex": 0,
+        "injectionType": "constructor"
+      },
+      "id": "1159932337"
+    },
     "1194720497": {
       "source": "543894580",
       "target": "-407945788",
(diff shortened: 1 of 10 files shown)


kamilmysliwiec:
could we add an e2e test for this?

yogeshwaran-c:
Added an e2e spec in `packages/apollo/tests/e2e/enum-union-directives.spec.ts`: it boots a federation app that uses `directives` on an enum, an enum value, and a union, then queries `_service.sdl` and asserts the tags round-trip into the subgraph SDL. Also refreshed the serialized-graph snapshot for the two new DI edges (`EnumDefinitionFactory`/`UnionDefinitionFactory` -> `AstDefinitionNodeFactory`).

kamilmysliwiec:
lgtm

PR #3951, opened 2026-04-22
feat(@nestjs/graphql): accept array of SDL strings in @Directive

yogeshwaran-c:
## Summary
- `@Directive` now also accepts a `string[]`, attaching every directive in a single decorator call.
- Each SDL string is validated individually and stored as its own metadata entry, so existing dedup and AST generation semantics are unchanged.

## Motivation
Stacking several federation/Contract directives requires repeating `@Directive(...)` for each SDL:

```ts
@Directive('@tag(name: "internal")')
@Directive('@tag(name: "restricted")')
@Directive('@inaccessible')
@ObjectType()
class User { ... }
```

This PR lets authors batch them together:

```ts
@Directive(['@tag(name: "internal")', '@tag(name: "restricted")', '@inaccessible'])
@ObjectType()
class User { ... }
```

The single-string form is fully preserved.

## Test plan
- [x] New spec `directive-multi-sdl.spec.ts` — class-level array, property-level array, and single-string forms all produce the expected AST directives.
- [x] `tests/schema-builder/**` continues to pass.

### DIFF
--- a/packages/graphql/lib/decorators/directive.decorator.ts
+++ b/packages/graphql/lib/decorators/directive.decorator.ts
@@ -4,29 +4,34 @@ import { LazyMetadataStorage } from '../schema-builder/storages/lazy-metadata.st
 import { TypeMetadataStorage } from '../schema-builder/storages/type-metadata.storage';
 
 /**
- * Adds a directive to specified field, type, or handler.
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
LGTM

PR #3962, opened 2026-04-28
fix(@nestjs/graphql): restore Timestamp scalar parsers in federation factory

yogeshwaran-c:
## Summary
- Extend the scalar restoration branch in `GraphQLFederationFactory.overrideOrExtendResolvers` to cover the built-in `Timestamp` scalar in addition to `DateTime`.
- Also copy `serialize` from the auto-generated scalar so the round-trip works regardless of which scalar shape the user picked.

## Bug
After `buildSubgraphSchema` re-parses the printed SDL, the resulting executable schema has scalars with default (no-op) parsers. The override step only restored `parseValue` / `parseLiteral` for the `DateTime` scalar — `DateTime` happened to render correctly anyway because the default serializer hands a `Date` straight to JSON which produces an ISO string, matching `GraphQLISODateTime.serialize`'s output.

When users opt into `dateScalarMode: 'timestamp'`, the schema uses `GraphQLTimestamp` whose `serialize` returns `value.getTime()`. Without the restoration, that scalar fell back to the default serializer in the federated schema, so resolvers that return a `Date` ended up with the raw ISO string in the response instead of a numeric epoch.

```ts
// before — only DateTime is restored
} else if (isScalarType(type) && type.name === 'DateTime') {
  type.parseLiteral = autoGeneratedScalar.parseLiteral;
  type.parseValue   = autoGeneratedScalar.parseValue;
}
```

## Fix
- Restore `Timestamp` alongside `DateTime`.
- Also restore `serialize` so the auto-generated scalar's full implementation comes through.

## Test plan
- [x] New e2e spec `packages/apollo/tests/e2e/federation-timestamp-scalar.spec.ts` boots an Apollo Federation app with `dateScalarMode: 'timestamp'`, returns a `Date(0)` from a resolver, and asserts the response value is the numeric `0`. Fails on master, passes after the fix.
- [x] Existing `tests/e2e/code-first-federation.spec.ts` continues to pass.


kamilmysliwiec:
lgtm

Issue #3764, opened 2025-12-12
Input type fields do not include extensions added by the @Extensions() decorator


thiagomini:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

When we use the @Extensions() decorator in an Input type and in one of its fields, the field-level extension is overridden by the class-level metadata.

Example Input:

```typescript
import { Extensions, Field, InputType } from '@nestjs/graphql';

@InputType()
@Extensions({ exampleExtension: 'exampleValue' })
export class CreateUserInput {
  @Extensions({ fieldLevelExtension: 123 })
  @Field()
  name: string;
}
```

Expected Schema Output:

```typescript
{
  name: 'CreateUserInput',
  extensions: { exampleExtension: 'exampleValue' },
  fields: [
    {
      name: 'name',
      type: 'String',
      extensions: { fieldLevelExtension: 123 }
    }
  ]
}

```

Actual Schema Output:

```typescript
{
  name: 'CreateUserInput',
  extensions: { exampleExtension: 'exampleValue' },
  fields: [
    {
      name: 'name',
      type: 'String',
      extensions: { exampleExtension: 'exampleValue' } // Incorrectly inherits from InputType
    }
  ]
}

```

### Minimum reproduction code

https://github.com/thiagomini/input-field-no-extension-bug

### Steps to reproduce

1. Clone the minimum repro
2. `npm run start:extensions`
3. See the error:

<img width="1521" height="812" alt="Image" src="https://github.com/user-attachments/assets/5268bbf4-23a9-4a27-84b0-eb4d1a9610c7" />

### Expected behavior

The field-level extension should not be overridden.

### Package version

13.2.0

### Graphql version

`graphql`: 16.12.0
`@apollo/server`: 5.2.0
`@as-integrations/express5`: 1.1.2

### NestJS version

11.1.9

### Node.js version

20.11.1

### In which operating systems have you tested?

- [ ] macOS
- [x] Windows
- [x] Linux

### Other

I know where the bug actually is: https://github.com/nestjs/graphql/blob/bfb75f2ba2701905974b1da21adeb4ef915c4ab3/packages/graphql/lib/schema-builder/factories/input-type-definition.factory.ts#L93

The input type definition factory should include the `property.extensions` in that line instead of `metadata.extensions`. I've confirmed that locally.

Issue #3328, opened 2024-10-10
@Extensions decorator don't work on @Field included in @InputType


Vittorio1970:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

**@ObjectType()**
```
export class SampleOutput {
  **@Extensions({ key: 'Value present in extensions member' })**
  @Field({
    nullable: true,
  })
  errorMessage: string;
}
```

**Result**
```
{
  name: "errorMessage",
  description: undefined,
  type: { ........ },
  args: [],
  resolve: (source, args, contextValue, info) => { ..... },
  subscribe: undefined,
  deprecationReason: undefined,
  extensions: {
    complexity: undefined,
    **key: "Value present in fieldconfig",**
  },
  astNode: undefined,
}
```
------------------------------------
**@InputType()**
```
export class SampleInput {
  @Extensions({ key: 'Value not present in field config' })
  @Field({
    nullable: false,
    defaultValue: false,
  })
  forceCreation: boolean;
}
```

**Result**

```
{
  name: "forceCreation",
  description: undefined,
  type: { .... },
  defaultValue: false,
  deprecationReason: undefined,
  **extensions: {
  },**
  astNode: undefined,
}
```

### Describe the solution you'd like

Same behavior for fields included in both InputType and ObjectType

### Teachability, documentation, adoption, migration strategy

.

### What is the motivation / use case for changing the behavior?

.

kamilmysliwiec:
Please provide a [minimum reproduction repository](https://minimum-reproduction.wtf/) (Git repository/[StackBlitz](https://stackblitz.com/fork/github/nestjs/typescript-starter)/[CodeSandbox](https://codesandbox.io/s/github/nestjs/typescript-starter/tree/master) project).

Vittorio1970:
https://codesandbox.io/p/devbox/friendly-newton-lnjrc3

Info from the schema handler

`[Object: null prototype] {
  OtypeMember: {
    name: 'OtypeMember',
    description: undefined,
    type: GraphQLScalarType {
      name: 'String',
      description: 'The `String` scalar type represents textual data, represented as UTF-8 character sequences. The String type is most often used by GraphQL to represent free-form human-readable text.',
      specifiedByURL: undefined,
      serialize: [Function: serialize],
      parseValue: [Function: parseValue],
      parseLiteral: [Function: parseLiteral],
      extensions: [Object: null prototype] {},
[...]

PR #3967, opened 2026-04-30
test(@nestjs/graphql): cover @extensions on @inputtype fields

maruthang:
## PR Checklist

(Test-only PR; no user-facing behavior is changed, so no documentation update is needed.)

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`@Extensions` applied to a `@Field` inside an `@InputType` was historically dropped from the resulting input field configuration, while the same decorator on `@ObjectType` fields worked correctly. The repo had no regression test pinning this parity, so the issue could resurface unnoticed.

Issue Number: #3328

## What is the new behavior?

The underlying source-code fix for #3328 already landed on `master` in commit [`1951a4c5`](https://github.com/nestjs/graphql/commit/1951a4c5) (merged under the duplicate report #3764), which made `InputTypeDefinitionFactory` read `field.extensions` instead of `metadata.extensions`. However, #3328 was never closed and no test guarded the fixed behavior.

This PR adds that missing regression coverage:

- New spec `packages/graphql/tests/schema-builder/factories/extensions-on-input-type.spec.ts` builds a tiny schema with both an `@ObjectType` and an `@InputType` whose fields carry `@Extensions({ role: 'admin' })`, then asserts the extensions object survives onto both the output and input field definitions in the generated `GraphQLSchema`.
- The test fails if `input-type-definition.factory.ts:101` is reverted to the pre-fix `metadata.extensions` access pattern, confirming it exercises the actual fix path rather than incidental behavior.

A second small commit (`chore(repo): allow lint-staged oxlint to ignore unmatched paths`) appends `--no-error-on-unmatched-pattern` to the `oxlint --fix` step in the `lint-staged` config. Without it, `oxlint` exits with a non-zero status whenever lint-staged hands it only files that match its `ignorePatterns` (e.g., commits that only touch `packages/**/tests/**`), which currently blocks any test-only commit. Happy to split this into a separate PR if maintainers prefer.

Note: I am an external contributor and cannot apply labels here, but `bug` / `needs triage` (and ideally a tests-only label) would be appropriate for triage.

- [ ] Yes
- [x] No

## Other information

- Verified locally with `yarn vitest run packages/graphql/tests/schema-builder/factories/extensions-on-input-type.spec.ts` — passes on `master`, fails when the source fix is reverted.
- No production code is modified by this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
   "lint-staged": {
     "*.ts": [
       "prettier --write",
-      "oxlint --fix"
+      "oxlint --fix --no-error-on-unmatched-pattern"
     ]
   },
   "repository": {
--- a/packages/graphql/tests/schema-builder/factories/extensions-on-input-type.spec.ts
+++ b/packages/graphql/tests/schema-builder/factories/extensions-on-input-type.spec.ts
@@ -1,0 +1,71 @@
+import { Test } from '@nestjs/testing';
+import { GraphQLInputObjectType, GraphQLObjectType } from 'graphql';
+import {
+  Args,
+  Extensions,
+  Field,
+  GraphQLSchemaBuilderModule,
+  GraphQLSchemaFactory,
+  InputType,
+  ObjectType,
+  Query,
+  Resolver,
+  TypeMetadataStorage,
+} from '../../../lib';
+
+@ObjectType()
+class OType {
+  @Field(() => String)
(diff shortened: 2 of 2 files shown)


kamilmysliwiec:
lgtm

PR #4054, opened 2026-07-01
feat(graphql): emit graphql descriptions as tsdoc comments

joelluijmes:
## What
Adds an opt-in `emitDescriptions` option to `DefinitionsGeneratorOptions`. When enabled, GraphQL schema descriptions (`"""..."""`) are emitted as TSDoc/JSDoc comments (`/** ... */`) on the generated TypeScript definitions.

Descriptions are carried through for:
- object / interface / input types
- fields (property signatures)
- resolver methods (root `Query`/`Mutation`/`Subscription`)
- scalars
- enums, enum members, and enum-as-type aliases
- unions

## Why
Today the schema-first definitions generator (`GraphQLDefinitionsFactory` → `GraphQLAstExplorer`) reads the AST but silently discards every node's `description`. This means the documentation authors write in their SDL never reaches the generated `.ts` file, so it's lost to editor tooltips and IntelliSense. This option surfaces that documentation in the generated types.

## Usage
```ts
const definitionsFactory = new GraphQLDefinitionsFactory();
await definitionsFactory.generate({
  typePaths: ['./src/**/*.graphql'],
  path: './src/graphql.ts',
  outputAs: 'interface',
  emitDescriptions: true,
});
```

Given:
```graphql
"""A feline animal."""
type Cat {
  """The name of the cat."""
  name: String!
}
```

produces:
```ts
/** A feline animal. */
export interface Cat {
    /** The name of the cat. */
    name: string;
}
```

## Backward compatibility
The option defaults to `false`, so existing generated output is unchanged unless a user opts in.

## Tests
Added specs in `graphql-ast.explorer.spec.ts` covering both the default (descriptions dropped) and enabled (descriptions emitted on types, fields, enums, and enum members) behavior. The full `@nestjs/graphql` suite passes (224 tests).

## Note
Per CONTRIBUTING this is a small, additive, opt-in feature so it's submitted directly as a PR. Happy to open a `[discussion]` issue first if maintainers prefer. Docs for the option can follow in the `nestjs/docs` repo if this is accepted.

### DIFF
--- a/packages/graphql/lib/graphql-ast.explorer.ts
+++ b/packages/graphql/lib/graphql-ast.explorer.ts
@@ -14,6 +14,7 @@ import {
   OperationTypeDefinitionNode,
   ScalarTypeDefinitionNode,
   ScalarTypeExtensionNode,
+  StringValueNode,
   TypeNode,
   TypeSystemDefinitionNode,
   TypeSystemExtensionNode,
@@ -25,6 +26,7 @@ import type {
   ClassDeclarationStructure,
   EnumDeclarationStructure,
   InterfaceDeclarationStructure,
+  JSDocStructure,
   MethodDeclarationStructure,
   MethodSignatureStructure,
   OptionalKind,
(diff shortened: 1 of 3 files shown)


joelluijmes:
@kamilmysliwiec what do you think of this improvement? Can we get this merged?

kamilmysliwiec:
lgtm

PR #4080, opened 2026-07-28
perf(graphql): avoid re-mapping resolver metadata per ancestor (90%+ performance gain on compileExtendedResolversMetadata function)

roypeled:
Disclosure - I used Claude Opus to investigate and fix this performance issue.

## PR Checklist

- [X] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

I went looking for why our schema build was slow, expecting to find it in our own code. Instead ~13s
of it was in `compileExtendedResolversMetadata`.

For every `(resolver, ancestor)` pair it re-`map`s all four metadata arrays — so
`O(resolvers × depth × handlers)`. On our biggest service (6,281 resolvers, ~13.5 avg depth) that's
~1.06B array-element visits, and nearly all of them copy an entry unchanged.

It's invisible until you have thousands of resolvers with deep inheritance; we only hit it because we
lean on generated base resolver classes.

Issue Number: N/A

## What is the new behavior?

Walk the prototype chains once, record which resolver inherits each base class, then apply it with one
`map` per array. Extracted into `ResolverImplementationMap`; `mergeParentResolverHandlers` and
`mergeParentFieldHandlers` are now unused and removed.

| resolvers | before | after |
| --------: | -----: | ----: |
| 750 | 8.0 ms | 0.6 ms |
| 3,600 | 202.0 ms | 2.2 ms |
| 6,000 | 538.7 ms | 4.0 ms |

Our own schema generation went from ~22.5s to ~7.6s per pass.

I wrote the new unit tests against the **existing** implementation first and confirmed they pass. I also diff-tested the two implementations
over 240 randomly generated hierarchies (~120k entries, shuffled registration order) - identical
everywhere except the change described below.

- [X] Yes
- [ ] No

In `mergeParentFieldHandlers`, the second `map`'s `metadata.target === parentClass` guard can never be
true — the call above it has just moved every such entry. I checked: 0 hits in 452,280 visits.

So the `objectTypeFn` fallback currently applies to *every* field resolver on *every* pair, and
converges on one shared value: the first non-throwing `typeFn` in registration order, which can come
from an unrelated hierarchy. Now it falls back to the `typeFn` of the resolver that actually owns it.

Only affects field resolvers whose `objectTypeFn` throws. Anyone depending on the old value was
depending on import order. **If you'd rather this PR were purely a perf change, say so and I'll
restore the old semantics.**

## Other information

A new warning was added when two resolvers on different branches extend the same base class
that declares handlers - only one can inherit them, and which one depends on import order. It's
lazily evaluated, and skips same-chain cases (otherwise every 3-level chain warns) and bases with no
handlers.

Happy to split this into three commits or drop the warning into its own PR — the perf change doesn't
depend on either of the others.

### DIFF
--- a/packages/graphql/lib/schema-builder/storages/resolver-implementation.map.ts
+++ b/packages/graphql/lib/schema-builder/storages/resolver-implementation.map.ts
@@ -1,0 +1,179 @@
+import { Logger } from '@nestjs/common';
+
+import {
+  BaseResolverMetadata,
+  FieldResolverMetadata,
+  ResolverClassMetadata,
+  ResolverTypeMetadata,
+} from '../metadata';
+import { isThrowing } from '../utils/is-throwing.util';
+
+/**
+ * Resolves which resolver class implements the handlers declared on a base class.
+ *
+ * A handler is recorded against the class it is declared on, but Nest instantiates the class
+ * registered as a provider. When a resolver extends a base class, every handler declared on that
+ * base therefore has to be reassigned to the derived class:
+ *
+ *   class BaseUserResolver { @Query(() => User) user() {} }    // target: BaseUserResolver
+ *   @Resolver() class UserResolver extends BaseUserResolver {} // the class Nest instantiates
+ *
+ * so `user` has to end up targeting `UserResolver`. Note that the base class need not be a resolver
+ * itself - `@Query` records whichever class the method is written on, decorated or not.
+ */
+export class ResolverImplementationMap {
+  private readonly resolverByBaseClass = new Map<
+    Function,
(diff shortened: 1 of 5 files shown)


kamilmysliwiec:
lgtm thanks @roypeled

PR #3973, opened 2026-05-05
chore: v14.0.0 release

kamilmysliwiec:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?
- [ ] Yes
- [ ] No

PR #4099, opened 2026-08-14
feat(@nestjs/graphql): allow custom WebSocket server in subscriptions config

### LINKED ISSUE #2653: Select websocket implementation in graphql-ws configuration

I cannot select another websocket implementation for subscriptions, and nest doesn't respect the selected websocket implementation from the adapter settings.

I was looking at this file https://github.com/nestjs/graphql/blob/master/packages/graphql/lib/services/gql-subscription.service.ts to see what websocket implementation would be used when enabling graphql-ws - it looks like it defaults to using the "ws" implementation from node (/lib/use/ws). 

It would be awesome if we could select which implementation was used, either through the registered adapter in Nest JS or just the ones that are provided by the author of Graphql-ws here: https://github.com/enisdenjo/graphql-ws/tree/master/src/use - which would allow selection of fastify, uWebsockets, or the default node implementation. 

_No response_

The node WS implementation is not as performant in my testing as these other libraries. Seems like it might require very little configuration changes to actually swap this library to another from his supported implementations?

Thank you for providing this library, Nest is like a dream come true :)

wasaybuilds:
## Description
This PR adds support for passing a custom WebSocket server instance via the `server` option in both `graphql-ws` and `subscriptions-transport-ws` configurations.

When provided, `@nestjs/graphql` will use the provided custom server instance instead of instantiating the default `ws` WebSocketServer.

Closes #2653

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation (types documented with JSDoc).
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

### DIFF
--- a/packages/graphql/lib/services/gql-subscription.service.ts
+++ b/packages/graphql/lib/services/gql-subscription.service.ts
@@ -28,6 +28,12 @@ export type GraphQLWsSubscriptionsConfig = Partial<
   >
 > & {
   path?: string;
+  /**
+   * A custom WebSocket server instance to use instead of the default
+   * `ws` WebSocketServer. When provided, the default WebSocket server
+   * will not be created for this protocol.
+   */
+  server?: WebSocketServer;
 };
 
 export type GraphQLSubscriptionTransportWsConfig = Partial<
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
it directly conflicts with changes made in https://github.com/nestjs/graphql/pull/3973 (next release)

Issue #3023, opened 2023-10-05
Provide way to delegate nested type resolution to resolver for that type


shohoney:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues _I found some related issues, but none of them were fully addressed.  I feel like my example is more concise on what the goals are, so I created it_

### Is your feature request related to a problem? Please describe it

When designing a graph that has deeply nested entities, there isn't a way to delegate resolution of those entities to their corresponding resolvers in a declarative way. Ideally, resolvers and resolver chains would be constructed in a way that delegates any entity resolution down to the resolver for the type, rather than needing to import a provider for the type to the parent.  This is particularly evident in federated graphs where our types are orphaned.  In this example, our graph doesn't actually define any root queries.  Our types are simply extensions of existing types.

Let's take the following example:
```typescript
@ObjectType()
export class Viewer { // This is a root type that's fetched via request context
  @Field()
  isAdmin: boolean;

  @Field()
  fullName: string;

  @Field(type => Company)
  company: Company;
  // ... many other fields.
}

@ObjectType()
export class Company {
  @Field()
  id: string;

  @Field()
  name: string;

  @Field(type => Configuration)
  someProductConfiguration: Configuration
}

@ObjectType()
export class Configuration {
  @Field(type => ID)
  id: string;

  @Field()
  name: string;

  // ... other fields
}

@Resolver(of => Viewer)
export class ViewerResolver {
  constructor(
    private readonly viewerLoader: Dataloader,
    private readonly companyLoader: Dataloader, // for the example below.  I ideally do not want to do this.
    private readonly productConfigurationLoader: Dataloader, // more dependency sprawl
  ) {}

  /**
   * there are a few field resolver examples here to show what's possible today.
   */
  @ResolveField(type => Company, { name: 'company' })
  async resolveCompanyBySpreadingParent(@Parent() viewer) {
    /**
     * This will execute the resolver of the corresponding type, but this entity cannot be nullable, otherwise we'll
     * get an error when trying to resolve fields on on that type which will naturally come back null since an entity
     * doesn't exist.  We get "cannot return null for non-null field..."
     */
    return { ...viewer }
  }

  @ResolveField(type => Company, { name: 'company' })
  async resolveCompanyWithProvider(@Parent() viewer) {
    /**
     * Great, we loaded a company.  However, there's a problem.  How do we resolve the `someProductConfiguration` field?
     * Once again, we have an entity that might look like
     * {
     *   id: string
[...]

Issue #2288, opened 2022-07-11
Implement loadeds for Mercurius driver


akaNightmare:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Mercurius provides the `loaders` options which allow avoiding 1:n problem, and it would be great to have this feature from the box.

### Describe the solution you'd like

A great solution exists in https://github.com/Davide-Gheri/nestjs-mercurius/ repo.

### What is the motivation / use case for changing the behavior?

No need to create own behaviour of the loaders for mercurius

kamilmysliwiec:
Just use `dataloader` package in combination with the request-scoped providers. A built-in "loaders" feature is not needed when you use Nest.

Please, use our [Discord](https://discord.gg/G7Qnnhy) channel (support) for such questions. We are using GitHub to track bugs, feature requests, and potential improvements.

Issue #94, opened 2018-12-17
Dataloader or another method to resolve n+1 problem


zerox12311:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[ ] Feature request
[X] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

I want to use dataloader to resolve n+1 problem , I has seen the issue, but I had no idea.

## Expected behavior

@kamilmysliwiec say will publish a small chapter in the docs about integration with DataLoader, but I haven't found it.....
Does anyone had resolve n+1 problem??

## Environment

<pre><code>
Nest version: 5.4.0
Nest graphql version: 5.4.0
apollo-server-express version: 2.3.1

 
For Tooling issues:
- Node version: 11.3.0
- Platform:  Windows

Others:

</code></pre>

rgolea:
@zerox12311 You can set it in context when you are using the GrapqhlModule or you can also generate it inside a middeware and generate it on every request.
Good luck with it!

biels:
I have implemented the dataloader in a project for the critical queries. I can share the code if you are interested. It uses interceptors and uses the context to provide the dataloaders. However, I think my implementation is far from ideal since it breaks most of nest and graphql semantics. It forces me to predict how resolvers will be executed in the queries that typically hit my server, which defeats the purpose of having a graphql api. 
I would like to (and I am willing to contribute with my experience) implement it natively into @nestjs/graphql with decorators or even by default. In my opinion, solving the n+1 problem is something that is needed in every graphql server, I cannot think of a use case where I would want to leave that out intentionally.

kamilmysliwiec:
Duplicate of #20

zerox12311:
> I have implemented the dataloader in a project for the critical queries. I can share the code if you are interested. It uses interceptors and uses the context to provide the dataloaders. However, I think my implementation is far from ideal since it breaks most of nest and graphql semantics. It forces me to predict how resolvers will be executed in the queries that typically hit my server, which defeats the purpose of having a graphql api.
> I would like to (and I am willing to contribute with my experience) implement it natively into @nestjs/graphql with decorators or even by default. In my opinion, solving the n+1 problem is something that is needed in every graphql server, I cannot think of a use case where I would want to leave that out intentionally.

Yes! I am interested in it!
Can you share you code ?

nguyenviettuananh:
@biels 
[...]

Issue #2, opened 2018-01-03
Batching and caching


marvinroger:
Hi!

I just discovered this framework and I have to say it’s awesome. Kudos!

The GraphQL module is great, but I did not find any information regarding how to do batching and caching, which is pretty required to avoid a big waste of resources (see https://github.com/facebook/dataloader).

Given the fact that resolvers are automatically mapped, I guess there’s currently no way to do that, right? An integration with dataloader would be awesome, if not mandatory for any medium to large application.

And, happy new year, by the way. ☺️

marvinroger:
Well, never mind! Resolvers are actually still functions, so we can use a dataloader there... sorry.

j:
@marvinroger any idea on clean ways to initialize a new DataLoader per request NOT in a resolver and using Nest's DI?

marvinroger:
I was thinking about this, but I did not have time to implement this in my side-project yet. I thought about a ‘dataloader’ module which would create every dataloaders. A middleware might be added right before the graphql one, it would call a method on the ‘dataloader’ module that would return a map of all dataloaders. Then, this map can be added as context for the graphql resolvers

Jonatthu:
Any example of how to use this?
@marvinroger

obedm503:
this is how I ended up implementing it
```ts
// app.module.ts
import {
  Module,
  MiddlewaresConsumer,
  NestModule,
  RequestMethod,
} from '@nestjs/common';
import { GraphQLModule, GraphQLFactory } from '@nestjs/graphql';
import { graphqlExpress } from 'apollo-server-express';
import * as DataLoader from 'dataloader';
import { Request } from 'express';                                                                                                                                

import { CatService } from './cat/cat.service';
import { CatResolver } from './cat/cat.resolver';

@Module({
  imports: [
    GraphQLModule,
  ],
  components: [
    CatService,
    CatResolver,
  ],
})
export class ApplicationModule implements NestModule {
  constructor(
    private readonly graphQLFactory: GraphQLFactory,
    private readonly catService: CatService,
  ) {}

  configure(consumer: MiddlewaresConsumer) {
    const typeDefs = this.graphQLFactory.mergeTypesByPaths('./**/*.gql');
    const schema = this.graphQLFactory.createSchema({ typeDefs });
    consumer
      .apply(
        graphqlExpress((req: Request) => {
          // this function is executed on every request
          // so a new catLoader is created each time

          const context = {
[...]

PR #4123, opened 2026-09-08
feat(@nestjs/graphql): add @BatchResolveField() for DataLoader-backed batching

Manuel-Antunes:
## Motivation

Spring for GraphQL ships [`@BatchMapping`](https://docs.spring.io/spring-graphql/reference/controllers.html): you annotate a controller method, it receives **every parent of the current execution layer at once**, and the framework wires the batching for you. The n+1 problem is solved declaratively, at the same level of abstraction you already write resolvers at.

Nest has no equivalent today. Solving n+1 means hand-rolling DataLoaders: instantiating them per request, threading them through the GraphQL context or making the resolver request-scoped, and re-implementing the positional key/value mapping in every field. That is boilerplate the framework is well positioned to absorb — and the ecosystem has repeatedly said so:

- [#2 Batching and caching](https://github.com/nestjs/graphql/issues/2) (2018)
- [#94 Dataloader or another method to resolve n+1 problem](https://github.com/nestjs/graphql/issues/94) (2018)
- [#2288 Implement loaders for Mercurius driver](https://github.com/nestjs/graphql/issues/2288) (2022)
- [#3023 Provide way to delegate nested type resolution to resolver for that type](https://github.com/nestjs/graphql/issues/3023) (2023)

The standing answer has been "use `dataloader` with request-scoped providers". That works, but it has a real cost: making a resolver request-scoped bubbles up its **entire dependency tree**, which is a performance trade-off users take on solely to get a per-request loader. `@BatchResolveField()` gives the same guarantee with no scope change at all — the loader lives in a `WeakMap` keyed by the GraphQL context object.

Worth noting: this has been asked for repeatedly but, as far as I can tell from searching the repo history, **never actually submitted as a PR**. This is an attempt to put an implementation on the table.

## What this is not

- **Not a breaking change.** Purely additive. No existing decorator, interface, or behavior changes. `@ResolveField()` is untouched; the batching branch in `ResolversExplorerService` only activates when the new metadata key is present.
- **Not a new dependency.** `dataloader` is an *optional* peer dependency, resolved through a dynamic `import()` exactly like the existing `ts-morph` integration. Users who never write a batch field never install it, and nothing in the published `.d.ts` references its types — the loader options are declared as a structural subset in `batch-loader-options.interface.ts`.
- **Not driver-specific.** The wiring lives in `@nestjs/graphql`, above the driver abstraction. E2E tests cover Apollo and Mercurius, code first and schema first.

## API

```ts
@Resolver(() => Post)
export class PostResolver {
  @BatchResolveField(() => [Comment])
  async comments(@Parent() posts: Post[]): Promise<Map<Post, Comment[]>> {
    const comments = await this.service.findByPostIds(posts.map((p) => p.id));
    return new Map(
      posts.map((post) => [
        post,
        comments.filter((c) => c.postId === post.id),
      ]),
    );
  }
}
```

The method may return:

- `Map<Parent, T>` — keyed by the parent objects (the Spring equivalent of relying on `equals`/`hashCode`)
- `Map<K, T>` with the `keyBy` option — because JS `Map` compares object keys by reference, `keyBy` is how you say "look this parent up by its id", which is what your repository has probably already grouped by
- `T[]` — one entry per parent, in order

Options are everything `@ResolveField()` accepts (`name`, `nullable`, `description`, `deprecationReason`, `complexity`, `middleware`) plus `keyBy` and `dataLoader` (forwarded to the underlying `DataLoader`: `maxBatchSize`, `cache`, `cacheKeyFn`, `batchScheduleFn`, `cacheMap`).

### One deliberate departure from Spring

Spring's `@BatchMapping` forbids field arguments, and sends you to `BatchLoaderRegistry` if you need them. Here, loaders are partitioned per `(request context, serialized arguments)`, so this batches correctly into two calls instead of erroring:

```graphql
{
  posts {
    published: comments(status: APPROVED) { text }
    pending: comments(status: PENDING) { text }
  }
}
```

## Implementation notes

- **Where it hooks in** (`resolvers-explorer.service.ts`): batching sits *between* the Nest resolver context and the field middleware. Field middleware therefore still runs once per field, while the method runs once per batch. Enhancers enabled via `fieldResolverEnhancers` sit inside the batch and run once per batch, with the array of parents as the execution-context root. Both behaviors are covered by tests.
- **Loader storage**: a per-registration `WeakMap<contextObject, Map<argsKey, DataLoader>>` in the closure, rather than writing onto the user's context object. No key collisions between resolvers, no trouble with frozen contexts, and loaders are garbage-collected with the request.
- **Type function**: required in code first (a batch method's reflected return type is `Map`/`Promise`, which says nothing about the schema field). The error is thrown at schema-build time, following the `@Query()` precedent, so schema first works with a bare `@BatchResolveField()`.
- **Missing package**: if `dataloader` is absent, the error is logged at boot and surfaced as a GraphQL error on the field, rather than failing silently.

## Tests

38 new tests, all green alongside the existing suite:

- `packages/graphql/tests/utils/batch-field-resolver.util.spec.ts` (17) — batching, per-request isolation, args partitioning, memoization, `maxBatchSize`, whole-batch and per-parent errors, fallbacks
- `packages/graphql/tests/decorators/batch-resolve-field.decorator.spec.ts` (7) — metadata and overloads
- `packages/apollo/tests/e2e/` (11) — code first, schema first, field middleware ordering, enhancer semantics, fast-path resolvers
- `packages/mercurius/tests/e2e/code-first-batch.spec.ts` (3) — driver independence

## Note on process

`CONTRIBUTING.md` asks for a `[discussion]` issue before a major feature. Given the four issues above already cover the request, I went straight to an implementation so there is something concrete to react to — but I am happy to move this to a discussion issue and iterate on the API (naming, `keyBy`, argument handling) before any of it is considered for merge.

Docs PR: nestjs/docs.nestjs.com#3527

### DIFF
--- a/packages/apollo/package.json
+++ b/packages/apollo/package.json
@@ -44,7 +44,8 @@
     "@nestjs/platform-fastify": "12.0.1",
     "@nestjs/testing": "12.0.1",
     "apollo-cache-inmemory": "1.6.6",
-    "apollo-client": "2.6.10"
+    "apollo-client": "2.6.10",
+    "dataloader": "2.2.3"
   },
   "dependencies": {
     "iterall": "1.3.0",
--- a/packages/apollo/tests/code-first-batch-enhancers/app.module.ts
+++ b/packages/apollo/tests/code-first-batch-enhancers/app.module.ts
@@ -1,0 +1,20 @@
+import { Module } from '@nestjs/common';
+import { GraphQLModule } from '@nestjs/graphql';
+import { ApolloDriver } from '../../lib/drivers/index.js';
+import { ApolloDriverConfig } from '../../lib/index.js';
+import { BlogService } from '../code-first-batch/blog.service.js';
+import { PostsResolver } from './posts.resolver.js';
+import { GuardCallsService, RecordingGuard } from './recording.guard.js';
+
+@Module({
+  imports: [
+    GraphQLModule.forRoot<ApolloDriverConfig>({
+      driver: ApolloDriver,
+      autoSchemaFile: true,
+      includeStacktraceInErrorResponses: false,
+      fieldResolverEnhancers: ['guards'],
+    }),
+  ],
+  providers: [BlogService, PostsResolver, GuardCallsService, RecordingGuard],
+})
+export class ApplicationModule {}
--- a/packages/apollo/tests/code-first-batch-enhancers/posts.resolver.ts
+++ b/packages/apollo/tests/code-first-batch-enhancers/posts.resolver.ts
@@ -1,0 +1,32 @@
+import { UseGuards } from '@nestjs/common';
+import { BatchResolveField, Parent, Query, Resolver } from '@nestjs/graphql';
+import { BlogService } from '../code-first-batch/blog.service.js';
+import { Author } from '../code-first-batch/models/author.js';
+import { Post } from '../code-first-batch/models/post.js';
+import { RecordingGuard } from './recording.guard.js';
+
+@Resolver(() => Post)
+export class PostsResolver {
+  constructor(private readonly blogService: BlogService) {}
+
+  @Query(() => [Post])
+  posts(): Post[] {
+    return this.blogService.findAllPosts();
+  }
+
+  @UseGuards(RecordingGuard)
+  @BatchResolveField(() => Author)
+  author(@Parent() posts: Post[]): Map<Post, Author> {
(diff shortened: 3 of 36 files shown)


kamilmysliwiec:
This PR was merged by accident during a batch merge that was meant for dependency updates only, so it's being reverted in #4148. Sorry for the noise. We'll revisit it separately with a proper review.

